PR #3, opened 2024-10-30
Many builtin implementations

utkn:
This PR aims to capture the semantics of builtin operations of Noir (i.e., which are referred to as *builtins*), and it makes the following additions:
- `Builtins/Arith.lean`: Arithmetic builtins for signed/unsigned integers and field elements
- `Builtins/Array.lean`: Array builtins (`len`, `as_slice`)
- `Builtins/Slice.lean`: Slice builtins, e.g., `push_back`, `pop_back`.
- `Builtins/BigInt.lean`: Builtins for `BigInt` types.
- `Builtins/Bit.lean`: Boolean logic and bitwise operation builtins for signed/unsigned integers.
- `Builtins/Cmp.lean`: Comparison builtins (`==`, `<`, and `>`)
- `Builtins/Field.lean`: Non-arithmetic field element (i.e., `Field` instance) builtins.
- `Builtins/Str.lean`: String builtins.
- `Builtins/Zeroed.lean`: The `zeroed` operation in Noir.
- `Tp.lean`: Definitions for arrays, strings, and `BigInt`s are added. Additionally, integers and unsigned integers are now represented by `BitVec`s.

The assumptions that we make are listed in the comments of the relevant builtins. However, there are some things to note:

- `Builtins/BigInt.lean`: The `BigInt` constructor `from_le_bytes(bytes: [u8], modulus: [u8])` is left as unimplemented. We need to learn how a `BigInt` is constructed from these values.
- `Builtins/Field.lean`: The conversion operations to/from unsigned/signed integers are left as unimplemented. It is not clear how the edge cases (e.g., overflows) are handled.
- `Builtins/Zeroed.lean`: We define the `zero` of a slice as an empty slice, and the reference type (`Ref`) as `0`. This needs to be checked.
- `Builtins/Comptime.lean`: This file contains the comptime/macro/reflection builtins and are unimplemented at this point.


Lampe/Lampe/Builtin/Field.lean
```diff
@@ -0,0 +1,168 @@
...
+  (fun h![a, w] _ => .reverse (extList
+    (withRadix 2 a.val (by tauto)) (w.toNat) 0))
+
+/--
+For a prime `p`, a field element `a : Fp p`, a radix `rad : U 32`, and an output length `w : U 32`,
+this builtin evaluates to a list of `w` little-endian radix `rad` digits representing `a`.
+We assume the following:
+- `a < rad^w` is not enforced, so the output may be truncated.
+- `1 < rad < 256` must hold, or this builtin throws an exception.
+In Noir, this builtin corresponds to `fn __to_le_radix(self, radix: u32, result_len: u32) -> [u8]` implemented for `Field`.
```
kustosz, line 42: This is not at all how it works in Noir, see docs in https://github.com/noir-lang/noir/blob/master/noir_stdlib/src/field/mod.nr

kustosz: Same with other decompositions in this file

kustosz: In fact, they cannot use the `newBuiltin` system – they are non-deterministic

Lampe/Lampe/Builtin/Field.lean
```diff
@@ -0,0 +1,168 @@
...
+
+In Noir, this builtin corresponds to `fn modulus_be_bytes() -> [u8]` implemented for `Field`.
+-/
+def fModBeBytes : Builtin := newBuiltin
+  [.field] (.slice (.u 8))
+  (@fun p _ => True)
+  (@fun p h![_] _ => .reverse (withRadix 256 p.val (by linarith)))
+
+/--
+Specs are not clear.
```
kustosz, line 129: What's unclear?

Lampe/Lampe/Builtin/Prelude.lean
```diff
@@ -0,0 +1,80 @@
...
+structure Builtin where
+  omni : Builtin.Omni
+  conseq : Builtin.omni_conseq omni
+  frame : Builtin.omni_frame omni
+
+end Lampe
+
+namespace Lampe.Builtin
+
+inductive builtinOmni
```
kustosz, line 37: This should be called something like `pureBuiltinOmni` or `functionalBuiltinOmni` – it's only valid for a subset of builtins that are deterministic partial functions wihout state access, so this very generic name is misleading

Lampe/Lampe/Builtin/Prelude.lean
```diff
@@ -0,0 +1,80 @@
...
+
+/--
+A `Builtin` definition.
+Takes a list of input types `argTps : List Tp`, an output type `outTp : Tp`, a predicate `pred` and evaluation function `comp`.
+For `args: HList (Tp.denote p) argTps`, we assume that a builtin function *fails* when `¬(pred args)`, and it *succeeds* when `pred args`.
+
+If the builtin succeeds, it evaluates to `some (comp args (h : pred))`.
+Otherwise, it evaluates to `none`.
+-/
+def newBuiltin
```
kustosz, line 59: same naming comment as `builtinOmni`

Lampe/Lampe/Builtin/Prelude.lean
```diff
@@ -0,0 +1,80 @@
+import Lampe.State
```
kustosz, line 1: This should be called `Basic.lean`. `Prelude` is a haskellism :D

Lampe/Lampe/Tp.lean
```diff
@@ -75,10 +79,14 @@ def Tp.denoteArgs : List Tp → Type
 @[reducible]
 def Tp.denote : Tp → Type
 | .u n => U n
+| .i n => I n
+| .bi => Int
 | .bool => Bool
 | .unit => Unit
+| .str => String
```
kustosz, line 86: AFAIK strings have type-level length in noir?

Lampe/Lampe/Tp.lean
```diff
@@ -75,10 +79,14 @@ def Tp.denoteArgs : List Tp → Type
...
 | .field => Fp P
 | .slice tp => List (denote tp)
+| .array tp _ => Array (denote tp)
```
kustosz, line 89: `Array` is not a good choice for this. It would be better served by a `Mathlib.Vector` or some custom type that carries the length evidence instead of forgetting it

Lampe/Lampe/Data/Integers.lean
```diff
@@ -2,8 +2,20 @@ import Mathlib
...
+instance : DecidableEq (I n) := inferInstanceAs (DecidableEq (BitVec _))
+
+instance : Repr (U n) where
+  reprPrec := fun a _ => a.toNat.repr
+
+instance : Repr (I n) where
+  reprPrec := fun a _ => a.toInt.repr
+
+/-- A proposition that checks whether the integer `val` can be represented by a `w`-width bit vector in 2s complement -/
+abbrev canContain (w : Nat) (val : Int) := val < 2^(w-1) ∧ val ≥ -2^(w-1)
```
kustosz, line 19: `canRepresent`?

Lampe/Lampe/Builtin/Zeroed.lean
```diff
@@ -0,0 +1,38 @@
+import Lampe.Builtin.Prelude
+
+namespace Lampe.Builtin
+open Lampe.Builtin
+
+/-- Returns a `zero` instance of a type -/
+def tpZeroed (tp : Tp) : (Tp.denote p tp) := match tp with
```
kustosz, line 7: This needs discussion with the noir team – I'm not sure if this is really correct

kustosz (changes requested):
Awesome stuff, but we need to be more careful in many places here

utkn:
- Arithmetic checks fixed.
- Bigint division now considers the divide-by-zero case.
- Definitions renamed as suggested.
- `assert` builtin now uses the pure builtin constructor.
- Memory builtins moved to `Memory.lean`
- Slice, string, and array builtins are converted to manual proofs to ensure that they are not parametrized over their type parameters.

Remaining questions:
- BigInt `from_le_bytes` implementation is left as unimplemented. This needs to be clarified. In general, we need to clarify how `BigInt` works.
- Field to/from int/uint conversions need to be clarified.
- `Zeroed` builtin needs to be clarified.
- Comptime builtins are left as a future task.

utkn:
- A new generic builtin constructor added.
- Replaced the manual proofs with new generic builtin constructors.
- Refactored the non-generic builtin constructor.

Lampe/Lampe/Builtin/Arith.lean
```diff
@@ -0,0 +1,217 @@
+import Lampe.Builtin.Basic
+namespace Lampe.Builtin
+open Lampe.Builtin
```
kustosz, line 3: no need to open your own namespace

Lampe/Lampe/Builtin/Basic.lean
```diff
@@ -0,0 +1,145 @@
...
+structure Builtin where
+  omni : Builtin.Omni
+  conseq : Builtin.omni_conseq omni
+  frame : Builtin.omni_frame omni
+
+end Lampe
+
+namespace Lampe.Builtin
+
+inductive pureBuiltinOmni
```
kustosz, line 37: This is isomorphic to `genericPureOmni (A := Unit)`. I suggest you rewrite it as such, so that we have fewer axioms. I.e. keep your `newPureBuiltin` but have it just delegate to `newGenericPureBuiltin`

Lampe/Lampe/Builtin/Cmp.lean
```diff
@@ -0,0 +1,143 @@
...
+  (fun tp h![a, b] => ⟨(tpEq tp).isSome,
+    fun h => ((tpEq tp).get h).beq a b⟩)
+
+/--
+Defines the equality comparison between unit values.
+We assume that `() == ()` always evaluates to `true`.
+
+In Noir, this builtin corresponds to `() == ()` implemented for `Unit`.
+-/
+def unitEq := newPureBuiltin
```
kustosz, line 52: all of these `xEq` definitions seem unnecessary after you've built `eq`? Probably can remove them

Lampe/Lampe/Builtin/Helpers.lean
```diff
@@ -0,0 +1,32 @@
...
+decreasing_by
+  rw [Nat.succ_eq_add_one, Nat.div_lt_iff_lt_mul]
+  rw [Nat.lt_mul_iff_one_lt_right]
+  tauto
+  exact Nat.succ_pos v'
+  rw [<-Nat.ne_zero_iff_zero_lt]
+  aesop
+
+-- [3, 2, 1]
+#eval (decomposeToRadix 10 123 (by linarith))
```
kustosz, line 21: again, convert #evals to `example`s

Lampe/Lampe/Builtin/Memory.lean
```diff
@@ -0,0 +1,81 @@
+import Lampe.Builtin.Basic
+namespace Lampe.Builtin
+open Lampe.Builtin
+
+inductive freshOmni : Omni where
```
kustosz, line 5: this doesn't belong in `memory` – probably can move back to `basic` for the lack of a better idea :D

Lampe/Lampe/Data/Integers.lean
```diff
@@ -2,8 +2,22 @@ import Mathlib
...
 
-instance : DecidableEq (U n) := inferInstanceAs (DecidableEq (Fin _))
+instance : DecidableEq (U n) := inferInstanceAs (DecidableEq (BitVec _))
+
+instance : DecidableEq (I n) := inferInstanceAs (DecidableEq (BitVec _))
+
+instance : Repr (U n) where
+  reprPrec := fun a _ => a.toNat.repr
+
+#eval ((BitVec.ofInt 8 (-128)).sdiv (BitVec.ofInt 8 (-1))).toInt
```
kustosz, line 15: another `example`

kustosz (approved):
Approving but please fix my comments before merging :)

PR #13, opened 2024-11-09
Introduction rules for builtins

utkn:
This PR adds the introduction rules for builtins. These reside in `Lampe/Hoare/Builtins.lean`.

Noteworthy changes:
- Generic `STHoarePureBuiltin` abbreviation added. Builtins that are defined with `newPureBuiltin` or `newGenericPureBuiltin` can use this to simplify the definition of the rule.
- Equality, addition, subtraction, multiplication and modulus builtins are converted to manual proofs. This is done because they are special cases that `newPureBuiltin` constructors do not work with. As a result, the references to these builtins in `Syntax.lean` are now correct and work with all possible types.
- Added `EqTp` and `ArithTp` types. Those types define the `Tp`s that can be compared with `==` and can accommodate the infix arithmetic operations `+`, `-`, `/`, `*` respectively.
- Arithmetic rules are added to the `steps` tactic. Note that now we may need to show that `ArithTp tp` and `EqTp tp` hold (see `Lampe.lean`).
- Documentation to `STHoare.lean` added that explains our interpretation of the Hoare triplets.

In the proofs, I took care to not use non-terminal `simp` or `simp_all` without an `only` directive.


utkn:
The equality and arithmetic builtins were converted back to typed automatic proofs (i.e., using `newGenericPureBuiltin` and `newGenericBuiltin` style). Accordingly, the intro rules are now typed, e.g., instead of `eq_intro` we have `fEq_intro`, `uEq_intro`, ... etc.

Lampe/Lampe/Hoare/SepTotal.lean
```diff
@@ -6,18 +6,65 @@ import Lampe.Hoare.Total
...
+Note that even if `e` terminates, it may evaluate to an error, e.g., division-by-zero.
+Accordingly, we interpret `{P} e {λv. Q v}` as follows:
+if `P` holds, then `e` terminates and it either (1) fails or (2) *successfully* evaluates to `v` such that `Q v` holds.
+This is logically equivalent to saying that if `P` holds, then `e` terminates and (`Q v` holds if `e` succeeds and evaluates to `v`).
+
+Hence, the triplets are *partial* with respect failure and *total* with respect to termination.
+
+An intuitive way of looking at this is thinking in terms of "knowledge discovery".
+For example, if the operation `a + b` succeeds, then we know that it evaluates to `v = a + b` **and** `a + b < 2^32`, i.e., no overflow has happened.
+Then, we would define the post-condition as `Q := λv. (v = a + b) ∧ (a + b < 2^32)`.
```
kustosz, line 28: I think this isn't valid lambda syntax – should use `\mapto` instead of `.`

Lampe/Lampe/Hoare/SepTotal.lean
```diff
@@ -6,18 +6,65 @@ import Lampe.Hoare.Total
...
+  (args : HList (Tp.denote p) (sgn a).fst) : Prop :=
+    STHoare p Γ ⟦⟧
+      (.call h![] (sgn a).fst (sgn a).snd (.builtin b) args)
+      (fun v => ∃h, v = (desc a (args)).snd h)
 
 namespace STHoare
 
+/--
+Separation logic frame rule.
+```
```
kustosz, line 46: I think these code blocks in all of this documentation are unhelpful – they verbatim the theorem statement, but with incorrect syntax. If anything, we may consider introducing this `{P} e {Q}` syntax as a macro for `STHoare`.

Lampe/Lampe/Hoare/SepTotal.lean
```diff
@@ -6,18 +6,65 @@ import Lampe.Hoare.Total
...
   unfold STHoare
   intro H'
   apply THoare.consequence ?_ (h_hoare (H ⋆ H')) ?_ <;> {
-    simp_all [SLP.star_assoc] -- [TODO] use SL-normalizer
+    simp_all only [SLP.star_assoc] -- [TODO] use SL-normalizer
     tauto
   }
 
+/--
+Separation logic consequence rule.
```
kustosz, line 60: Again, all the documentation following this can be removed. "Repeat function name and then repeat the statement"  is not helpful documentation.

Good documentation for this would mention why one may want to use this, when not to use it etc. I'm not saying you should spend time writing good documentation here, but you shouldn't commit bad one :P

Lampe/Lampe/Hoare/Total.lean
```diff
@@ -5,6 +5,12 @@ import Lampe.SeparationLogic
 
 namespace Lampe
 
+/--
+  A Hoare triplet.
```
kustosz, line 9: `triple`

Lampe/Lampe/Hoare/Total.lean
```diff
@@ -5,6 +5,12 @@ import Lampe.SeparationLogic
...
+  ```
+  {P} e {Q} ≡ ∀st, P st → (e ⇓ Q)
```
kustosz, line 11: This would be more useful if you used words here. Or remove :D

kustosz (approved):
Some changes needed but I don't need to re-review

PR #16, opened 2024-11-15
Lambda support

### LINKED ISSUE #6: Lambdas
Lambda functions should be supported by the tool. This needs a level of indirection to avoid non-termination issues. This can be achieved by shoving their runtime reps (which is going to be `HList Tp.denote argTypes -> Expr Tp.denote outType`) onto the heap and pass them around in references.

AST, syntax and semantics need to be extended to support lambda calls.

A good test case for this would be a module that
1. Defines `Slice.reduce`
2. Calls it with e.g. a lambda that adds numbers
3. The postcondition that this is a sum of all items in the slice should be provable with all hoare-logic steps skipped by automation (`steps` and `sl`)

utkn:
This PR aims to add lambda support. To that end, the following changes have been made:
- A new typeclass `LawfulHeap` has been added. This denotes the types that can be argued with separation logic predicates. Implementors must define the operations (union and disjoint) and prove some fundamental theorems about them. `SLP` now works over all types that implement `LawfulHeap`. `State _` has been replaced with two types that implement `LawfulHeap`: `ValHeap` and `State` where `ValHeap` is the old state and `State` is a wrapper around `ValHeap` that adds a set of lambdas. We can construct lambda singletons with the `[r ↣ l]` which means `r` refers to the lambda `l`. This is a big change, and required many changes across the codebase.
- **AST**: A new structure `Lambda` is introduced. A lambda expression `Expr.lam` is introduced. Moreover, `FunctionIdent` is extended to support references to lambda functions in the environment. Accordingly, lambda calls can be represented with `Expr.call`s.
- **Syntax**: New syntax for creating and calling lambdas have been introduced. Lambdas can be constructed as 
`let add = |a, b| : u8, u8 -> u8 { ... }`, and they can be called with `^add(4 : u8, 7 : u8)`
- **Semantics**: Two new Omni rules (one for creating a lambda and one for calling lambdas) were added. A new theorem `nested_triple` is added that can be used by the steps tactic to deconstruct the goals of form `P {e} Q` with introduction rules of form `{P1} e1 {Q} -> {P1 * H} e2 {Q * H}`, like `callLambda_intro`. Corresponding introduction rules were added and integrated into the `steps` tactic.

Closes #6


Lampe/Lampe/Semantics.lean
```diff
@@ -34,49 +34,57 @@ inductive TraitResolution (Γ : Env): TraitImplRef → List (Ident × Function)
...
   TraitResolution Γ ref (impl.impl implGenerics)
 
-inductive Omni : Env → State P → Expr (Tp.denote P) tp → (Option (State P × tp.denote P) → Prop) → Prop where
-| litField : Q (some (st, n)) → Omni Γ st (.lit .field n) Q
-| litFalse : Q (some (st, false)) → Omni Γ st (.lit .bool 0) Q
-| litTrue : Q (some (st, true)) → Omni Γ st (.lit .bool 1) Q
-| litU {Q} : Q (some (st, ↑n)) → Omni Γ st (.lit (.u s) n) Q
-| var : Q (some (st, v)) → Omni Γ st (.var v) Q
-| skip : Q (some (st, ())) → Omni Γ st (.skip) Q
+inductive Omni : (p : Prime) → Env → State p → Expr (Tp.denote p) tp → (Option (State p × Tp.denote p tp) → Prop) → Prop where
```
kustosz, line 37: what was wrong with the variable on line 11?

Lampe/Lampe/Semantics.lean
```diff
@@ -154,6 +206,48 @@ theorem Omni.frame {p Γ tp st₁ st₂} {e : Expr (Tp.denote p) tp} {Q}:
...
+    obtain ⟨_, hd⟩ := h
+    constructor
+    intros
+    simp only
+    rename Lambdas => lmbs
+    rename ValHeap _ => vh
+    rename Ref => r
+    generalize hL : (⟨_, _, _, _⟩ : Lambda) = lambda
+    have hi : r ∉ lmbs ∧ r ∉ st₂.lambdas := by aesop
+    have hd₁ : Finmap.Disjoint lmbs (Finmap.singleton r lambda) := by
```
kustosz, line 229: All of this set-fu needs to become lemmas, way too much repetition in this code. Let's rush features first, but please open an issue to fix these up when we have time

Lampe/Lampe/SeparationLogic/LawfulHeap.lean
```diff
@@ -0,0 +1,75 @@
...
+  union_assoc {s₁ s₂ s₃ : α} : union (union s₁ s₂) s₃ = union s₁ (union s₂ s₃)
+  disjoint_symm_iff {a b : α} : disjoint a b ↔ disjoint b a
+  union_comm_of_disjoint {s₁ s₂ : α} : disjoint s₁ s₂ → union s₁ s₂ = union s₂ s₁
+  disjoint_union_left (x y z : α) : disjoint (union x y) z ↔ disjoint x z ∧ disjoint y z
+  disjoint_empty (x : α) : disjoint empty x
+
+instance [LawfulHeap α] : Union α := ⟨LawfulHeap.union⟩
+instance [LawfulHeap α] : EmptyCollection α := ⟨LawfulHeap.empty⟩
+
+theorem LawfulHeap_union_empty_commutative [LawfulHeap α] {a : α} :
```
kustosz, line 20: All of these should become namespaces instead of prefixes, like `LawfulHeap.union_empty_commutative`

kustosz: Also make it `comm` nobody has time to write `commutative` :D

Lampe/Lampe/SeparationLogic/State.lean
```diff
@@ -0,0 +1,116 @@
...
+def State.valSingleton (r : Ref) (v : AnyValue p) : SLP (State p) :=
+  fun st => st.vals = Finmap.singleton r v
+
+notation:max "[" l " ↦ " r "]" => State.valSingleton l r
+
+@[reducible]
+def State.lmbSingleton (r : Ref) (v : Lambda) : SLP (State p) :=
+  fun st => st.lambdas = Finmap.singleton r v
+
+notation:max "[" l " ↣ " r "]" => State.lmbSingleton l r
```
kustosz, line 91: Not a fan of this notation, these arrows are hard to see. I propose `[ \lam l \map r]`. This way we keep the arrow and clearly say "this is a lambda sir"

Lampe/Lampe/SeparationLogic/ValHeap.lean
```diff
@@ -0,0 +1,47 @@
...
+  disjoint := fun a b => a.Disjoint b
+  empty := ∅
+  union_empty := Finmap.union_empty
+  union_assoc := Finmap.union_assoc
+  disjoint_symm_iff := by tauto
+  union_comm_of_disjoint := Finmap.union_comm_of_disjoint
+  disjoint_union_left := Finmap.disjoint_union_left
+  disjoint_empty := Finmap.disjoint_empty
+
+-- def ValHeap.singleton (r : Ref) (v : AnyValue p) : SLP (ValHeap p) := fun st => st = Finmap.singleton r v
```
kustosz, line 43: rm

kustosz (changes requested):
Some changes needed, but this awesome! I'm glad this works and turned out so nice

Lampe/Lampe.lean
```diff
@@ -89,7 +90,27 @@ example {p Γ x y}: STHoare p Γ ⟦⟧ (simple_if_else.fn.body (Tp.denote p) h!
   . simp only [decide_True, exists_const]
     sl
     contradiction
-  . simp_all
+  . aesop
+
+nr_def simple_lambda<>(x : Field, y : Field) -> Field {
+  let add = |a, b|: Field, Field -> Field { #add(a, b) : Field };
```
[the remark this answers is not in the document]
utkn: I had tried this but couldn't get it to work. But now I was able to do it.

Lampe/Lampe/Semantics.lean
utkn: Issue opened #19

kustosz (approved):

PR #21, opened 2024-12-04
Proof simplifications & additional lemmas

### LINKED ISSUE #19: Additional lemmas
We need to define some lemmas for `LawfulHeap` and `Finmap`s so that the proofs get simpler.

utkn:
Closes #19


kustosz:
Doesn't compile and is stale, closing

PR #22, opened 2024-12-07
Struct support

### LINKED ISSUE #11: Revive struct support
Structs are currently commented out. We should bring them back. Struct creation should be a builtin parametrized by the struct definition. Similarly projections should be handled by builtins (parametrized by the `Member` structure)

utkn:
This PR aims to add struct support and it closes #11. To that end, we make the following additions:

- Replaced `Tp.struct` with `Tp.tuple` which accepts an optional name.  This allows us to reason about structs and tuples with a single underlying type. Structs are represented as named tuples.
- Added the builtins `mkStruct` and `projectTuple`. These builtins create a named tuple and get the appropriate element of a tuple respectively. Note that `projectTuple` is parametrized over the "accessor" which is a `Fin _` and it denotes the index of the projection, e.g., `projectTuple $ Fin.mk 2 _` is a builtin that returns the third element of a tuple.
- Introduction rules `mkStruct_intro` and `projectTuple_intro` are added.
- Syntax support is added. See `Lampe.lean` for their usage.


utkn:
- Struct names are not removed to not mess with the trait resolution
- `Member` struct is now utilized as opposed to `Fin _`
- On a field `a` of a struct `Pair` we define `Pair#a` which is of type `Member _ _ `. The builtin that gets that field is then `projectTuple Pair#a`.

kustosz (approved):

PR #24, opened 2024-12-13
Extractor improvements

utkn:
This PR aims to update the extractor so that it outputs valid Lean definitions.


mpenciak (approved):
Looks good!

PR #29, opened 2025-01-01
Higher order functions

utkn:
This PR aims to add higher order function support. 
- A new type is added `Tp.fn`
- Functions are represented by `FuncRef`s
- A new expression is added `Expr.fn` for function idents
- The call expression `Expr.call` is updated to work with a `Tp.fn`
- Call intro rules `callDecl_intro`, `callLambda_intro`, `callTrait_intro` are updated
- Syntax is updated

### DIFF
--- a/Lampe/Lampe.lean
+++ b/Lampe/Lampe.lean
@@ -146,7 +146,8 @@ example {p} {arg : Tp.denote p Tp.field} :
   simp only [simple_trait_call]
   steps
   . apply STHoare.callTrait_intro
-    apply SLP.ent_star_top
+    sl
+    tauto
     try_impls_all [] simpleTraitEnv
     all_goals tauto
     simp only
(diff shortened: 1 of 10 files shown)


Lampe/Lampe.lean
```diff
@@ -9,8 +9,8 @@ nr_def simple_muts<>(x : Field) -> Field {
   y
 }
 
-example : STHoare p Γ ⟦⟧ (simple_muts.fn.body _ h![] |>.body h![x]) fun v => v = x := by
-  simp only [simple_muts]
+example : STHoare p Γ ⟦⟧ («fn#simple_muts».fn.body _ h![] |>.body h![x]) fun v => v = x := by
```
kustosz, line 12: this name change is painful, makes it much worse for the user, can we work around this?

kustosz (commented):
This is generally awesome, but can we get back to the original naming convention?

kustosz (approved):

PR #36, opened 2025-01-13
String literals

mpenciak:
This PR adds string literals. A follow up PR will add format strings.

It's ready for review, but the `trace`s in `Tactic/SeparationLogic.lean` should be deleted after I do a little more experimenting.

Something to consider: Getting rid of `FixedLenStr` completely. It basically plays the role of `List.Vector Char (n : Nat)` wherever it's being used, and the additional `.toVector`, `.fromVector` hassle makes it somewhat inconvenient to use.


Lampe/Lampe/Ast.lean
```diff
@@ -21,7 +22,8 @@ structure TraitMethodImplRef where
   method : Ident
 
 inductive Expr (rep : Tp → Type) : Tp → Type where
-| lit : (tp : Tp) → Nat → Expr rep tp
+| litNum : (tp : Tp) → Nat → Expr rep tp
+| litStr : (len : U 32) → FixedLenStr len.toNat → Expr rep (.str len)
```
utkn, line 26: What happens if we replace `FixedLenStr len.toNat` with `String` here? I think we should be able to get rid of `FixedLenStr`. I don't see a reason why we have to carry the length proof along with the string.

utkn (approved):
Thank you, this looks great! I have added some comments. One is regarding to your proposal on removing `FixedLenStr`, which I agree with.

Lampe/Lampe/Ast.lean
mpenciak: I think the issue is that when we get to working through the semantics, there is no relation between the `len : U 32` and the length of the raw `String`.

utkn: You're right, I see it now.

utkn (approved):

PR #37, opened 2025-01-14
Update Noir version to 1.0.0-beta.1

### LINKED ISSUE #35: Update targeted Noir version
We need to update the Noir version we are building against.

utkn:
This PR updates the Noir version to `1.0.0-beta.1`, and it closes #35.

- The new Noir repository resides in https://github.com/reilabs/noir/tree/lampe-v1.0.0-beta.1. Accordingly, the CI is updated to checkout the branch `lampe-v1.0.0-beta.1`. Readme is also updated to reflect this change.
- I use `RUST_MIN_STACK_SIZE` to set the stack size of spawned threads to 16MBs, otherwise CI step `cargo test` fails.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,12 +28,12 @@ jobs:
             uses: actions/checkout@v4
             with:
               path: lampe
-          - name: "Checkout reilabs/noir (branch: lampe)"
+          - name: "Checkout utkn/noir (branch: lampe)"
             uses: actions/checkout@v4
             with:
               path: noir
               ref: lampe
-              repository: reilabs/noir
+              repository: utkn/noir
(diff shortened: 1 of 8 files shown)


mpenciak:
Looks good to me! Only one question I have is: Why are we relying on a local version of `noir` and not just pointing to our fork + branch? I see the benefit when it comes to development, but I think the README instructions
and CI could be a bit easier if we did it that way.

mpenciak (approved):

utkn:
I can't see a reason why we shouldn't proceed with a remote repo as you suggested. Please have a look at the latest commit, and let me know if this is what you meant.

mpenciak (approved):
Yeah, that's what I had in mind. Looks good!

PR #39, opened 2025-01-18
TraitAsType support

utkn:


Lampe/Lampe/Builtin/Basic.lean
```diff
@@ -41,16 +41,40 @@ def HList.toVec (l : HList rep tps) (h_same : tps = List.replicate n tp) : List.
...
-| .nil => ()
-| .cons arg args => ⟨arg, HList.toTuple args name⟩
+def HList.toCTps {tps : List Tp} {tps' : List CTp}
+  (l : HList (Tp.denote p) tps)
+  (h_same : tps = ↑tps') : HList (CTp.denote p) tps' :=
+    match tps, l with
+    | [], .nil => match tps' with | [] => HList.nil
+    | tp :: tps, .cons x xs => match tps' with
+      | tp' :: tps' => HList.cons
+        (by
```
mpenciak, line 51: Going into tactics mode in this definition has the potential to make it difficult to reason about. Is there any way to avoid this?

For what it's worth, I played around with this a bit, and wasn't able to figure it out so the answer could totally be "no"... So if it takes too much time then we can ignore it for now and get back to it.

mpenciak (approved):
I left a couple small comments, but otherwise this looks good!

Sorry about the delay in review, I took some time to understand the stuff that's happening in Lampe/Hoare/* and this was quite helpful.

Lampe/Lampe/Builtin/Basic.lean
mpenciak: Actually I think I figured it out:
```lean
      | tp' :: tps' =>
        have h_same : tp' = tp := by simp_all
        have h_same' : Tp.denote p tp = CTp.denote p tp' := by rw [←h_same]
        have h_sames : tps = tps' := by simp_all
        HList.cons (h_same' ▸ x) (HList.toCTps xs (by tauto))
```
Looking at the outputed expr, this doesn't actually make things any better than before... so it's not important to go with the suggestion or not

utkn:
Closing for #52 that replaces this one.

PR #40, opened 2025-01-20
Format Strings (kind of)

mpenciak:
This PR adds support for Format Strings.

So far the extractor, syntax, and type definitions are defined. Not defined is the concrete type representing format strings, and their semantics.

In the process of writing builtin support I was running into universe issues with some of the naive implementation attempts. That combined with the fact that Noir has minimal support for format strings (as far as I can intuit from my experiments), I decided to stub out the definition and get this PR ready for review sooner rather than later.

I'll make sure to add an issue for full format string semantics eventually.

### DIFF
--- a/Lampe/Lampe/Syntax.lean
+++ b/Lampe/Lampe/Syntax.lean
@@ -43,8 +43,7 @@ syntax ident ":" nr_type : nr_param_decl
 
 syntax num ":" nr_type : nr_expr -- Numeric literal
 syntax str : nr_expr -- String literal
-syntax "#format(" str "," nr_fmtstr_part,* ")" : nr_expr -- Foramt string
-syntax "("nr_expr ":" nr_type ")" : nr_fmtstr_part
+syntax "#format(" str "," nr_expr,* ")" : nr_expr -- Foramt string
 syntax "#unit" : nr_expr -- Unit literal
 syntax ident : nr_expr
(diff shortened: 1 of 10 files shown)


Lampe/Lampe/Syntax.lean
```diff
@@ -427,6 +439,11 @@ partial def mkExpr [MonadSyntax m] (e : TSyntax `nr_expr) (vname : Option Lean.I
 | `(nr_expr| `( $args,* )) => do
   mkArgs args.getElems.toList fun argVals => do
     wrapSimple (←`(Expr.mkTuple none $(←mkHListLit argVals))) vname k
+| `(nr_expr| #format($s:str, $args,*) ) => do
+  let fmtStrParts ← (args.getElems.toList.mapM mkFmtStrPart)
+  let (partVal, partType) := fmtStrParts.unzip
+  mkArgs partVal fun _argVals => do -- TODO: Eventually we'll need the `argVals` when constructing the fmtStr
+    wrapSimple (←`(Expr.fmtStr (String.length $s) $(← mkListLit partType) (Unit.unit))) vname k
```
utkn, line 446: At this point, we should be able to directly get the type of the argument values like `let argTps := argVals.map fun arg => typeOf arg` where `def typeOf (rep tp) = tp`. So I think we can remove the need for specifying the type of the arguments at the syntax level with `nr_fmtstr_part` syntax.

utkn (approved):
Looks great, thanks!

PR #41, opened 2025-01-22
Pattern support

utkn:
This PR aims to add pattern support. Patterns are only handled at the extractor side -- all the patterns are converted to either `let` or `let mut` bindings. 

For example, the statement 
```rust
let (mut x, y) = (1, 2)
```

is converted to

```
let param#0 = `(1, 2);
let mut x = param#0.0;
let y = param#0.1;
```


mpenciak (approved):
Looks good!

PR #52, opened 2025-01-24
TraitAsType support

utkn:
This PR aims to add `TraitAsType` support. They are handled purely at the extractor side. 

If they occur in parameter types, they are replaced by type variables. If they occur in the return type, they are replaced by the concrete type that the function body returns.

I also added a placeholder type at the Lean side, which can be invoked with `_` in the Lean definitions. This might be useful in the future.


mpenciak (approved):
Looks like a much simpler solution than #39!

Issue #55, opened 2025-01-30
Demo


mpenciak:
The goal is to make a Merkle tree library in Noir that is generic in the hash function, extract the code, and provide proofs of the specification by interfacing with the Merkle tree library in [proven-zk](https://github.com/reilabs/proven-zk).

We should have a hash function instantiated using the Skyscraper hash (https://github.com/reilabs/whir/blob/skyscraper/src/skyscraper.rs and https://eprint.iacr.org/2025/058)

mpenciak:
Done in #71

PR #56, opened 2025-02-01
Noir Merkle Tree implementation

utkn:
This PR adds a Merkle tree implementation in Noir to be used for the demo #55.

For now, this simply adds a Merkle tree implementation in `test_fixtures/src/mtree.nr`. Later, this needs to be combined with the hash implementation to construct the full demo.

I have encountered compiler bugs when I wanted to implement `root` and `set_leaf` by recursion. Hence, these use an iterative implementation that might be harder to show equivalence with the [target definition](https://github.com/reilabs/proven-zk/blob/main/ProvenZk/Merkle.lean).


src/lean/mod.rs
```diff
@@ -1279,10 +1278,10 @@ impl LeanEmitter {
             }
             HirStatement::For(fors) => {
                 let loop_var = self.context.def_interner.definition_name(fors.identifier.id);
+                let loop_var = &syntax::expr::format_var_ident(loop_var);
```
mpenciak, line 1281: What was going wrong that needed this change? Just curious for my understanding (and similarly in `src/lean/pattern.rs`

src/lean/pattern.rs
```diff
@@ -60,59 +67,83 @@ fn parse_pattern(pat: &HirPattern, ctx: &mut PatCtx) -> Vec<PatRes> {
...
-                PatType::Struct { struct_type, field } => {
-                    let struct_name = emitter.emit_fully_qualified_type(&struct_type, emitter_ctx);
-                    rhs = super::syntax::expr::format_member_access(&struct_name, &rhs, &field.to_string());
-                },
+    parse_pattern(pat, &mut ctx)
+        .into_iter()
+        .map(|pat_res| {
+            let PatRes(id, ctx) = pat_res;
+            let lhs = emitter.context.def_interner.definition_name(id.id).to_string();
+            let lhs = super::syntax::expr::format_var_ident(&lhs);
```
mpenciak, line 124: This is the line I referred to above.

src/lean/mod.rs
```diff
@@ -1293,8 +1292,14 @@ impl LeanEmitter {
...
         };
-
-        Ok(format!("{result};"))
+        // Append a semicolon to the statement if it doesn't already end with one.
+        // We emit some statements already with a semicolon, so we don't want to add another.
+        // [TODO] maybe fix this for consistency
+        if result.ends_with(";") {
+            Ok(result)
+        } else {
+            Ok(format!("{result};"))
```
mpenciak, line 1301: Wouldn't this add a `;` to the end of return statements as well?

mpenciak (commented):
Added some comments... I'm thinking maybe we can wait to merge until after #59, #58, and then I can put in my hash implementation and we can merge it all at once.

src/lean/mod.rs
utkn: At this point, Noir doesn't have early returns. The last expression is always the return value of a function.

kustosz:
Got a better version on a branch, closing

PR #69, opened 2025-03-10
minimal example

mpenciak:
This PR adds a minimal example to show off Lampe usage.


Examples/example1/lakefile.toml
```diff
@@ -0,0 +1,12 @@
+name = "example1"
+version = "0.1.0"
+defaultTargets = ["Example1"]
+
+[[lean_lib]]
+name = "Example1"
+
+[[require]]
+name = "Lampe"
```
kustosz, line 9: this should point to the directory, not the repo – so that we can use this for working on Lampe as well

kustosz (approved):
Awesome just one thing I'd like to change

PR #71, opened 2025-03-12
Skyscraper example

mpenciak:
This PR adds the skyscraper example from the work in progress [here](https://github.com/mpenciak/nr_skyscraper).

It also renames the `Example1` example to `Merkle`, and deletes a spurious folder generated by Nargo in the root


Examples/Merkle/Merkle/Spec.lean
```diff
@@ -0,0 +1,434 @@
...
+
+lemma SkyscraperHash_correct: STHoare lp env ⟦⟧ (Expr.call [Tp.field, Tp.field] Tp.field
+          (FuncRef.trait (some $ «struct#Skyscraper».tp h![]) "BinaryHasher" [Kind.type] (HList.cons Tp.field HList.nil) "hash" [] HList.nil) h![a,b]) (fun v => v = Ref.State.compress ⟨[a, b], rfl⟩) := by
+  enter_trait [] env
+  steps [permute_intro]
+  casesm*∃_,_
+  subst_vars
+  congr 1
+
+theorem main_correct [Fact (CollisionResistant Ref.State.compress)] {tree : MerkleTree (Fp lp) Ref.State.compress 4}:
```
kustosz, line 420: Why only depth 4? Let's do 32 or something like that

Examples/Merkle/Prover.toml
```diff
@@ -0,0 +1,4 @@
+root = "0x250c97c9d3e44e27a4781a76aa934c1c554a4829b8d4bdc9fb5818f71875944b"
```
kustosz, line 1: I think this file is not needed?

Examples/test.sh
```diff
@@ -0,0 +1,87 @@
...
+	$CLI --root $dir"/src" --out-file $extracted
+
+	lake exe cache get
+	lake build
+
+	rm $extracted
+	mv $extracted_temp $extracted
+done
+
+# LEAN_ROOT=$PROJECT_ROOT/Lampe/
```
kustosz, line 30: Remove the commented out stuff

kustosz (approved):
Minor change requests, but this is awesome!

Examples/Merkle/Merkle/Spec.lean
mpenciak: I set it to 4 so I could actually run some examples locally (32 was taking way too long). I'll re-set it to 32 for the example though

PR #75, opened 2025-03-20
Field generics

mpenciak:
This PR adds generic field elements.

Small notes associated with the PR:

* Notation for unsigned int generics was changed from just the bit-width `w` to `type_u w`
* Added notation for generic field elements as `type_fp`
* Numeric consts now have different notation depending on the type. `f@A` for `A : type_fp` and `u@B` for `B : type_u w`
* I also ran `cargo fmt` to normalize the rust formatting and prevent errant formatting diffs in future PRs (it may be worth adding this to CI in the future)

Small TODO: 
* The proof of the final lemma in `Lampe/Tests/FieldGenerics.lean` is a tad bit messy (`steps` was not able to finish the proof automatically), so it may be worth looking at why.

### DIFF
--- a/Examples/Merkle/Merkle/Extracted.lean
+++ b/Examples/Merkle/Merkle/Extracted.lean
@@ -8,12 +8,12 @@ nr_struct_def Skyscraper<> {
 
 }
 
-nr_def «SIGMA»<>() -> Field {
-    9915499612839321149637521777990102151350674507940716049588462388200839649614 : Field
-}
-
 nr_def «RC»<>() -> [Field; 8] {
     [-4058822530962036113558957735524994411356374024839875405476791844324326516925 : Field, 5852100059362614845584985098022261541909346143980691326489891671321030921585 : Field, -4840154698573742532565501789862255731956493498174317200418381990571919688651 : Field, 71577923540621522166602308362662170286605786204339342029375621502658138039 : Field, 1630526119629192105940988602003704216811347521589219909349181656165466494167 : Field, 7807402158218786806372091124904574238561123446618083586948014838053032654983 : Field, -8558681900379240296346816806663462402801546068866479372657894196934284905006 : Field, -4916733727805245440019875123169648108733681133486378553671899463457684353318 : Field]
+}
+
+nr_def «SIGMA»<>() -> Field {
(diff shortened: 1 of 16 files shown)


Examples/Merkle/Merkle/Extracted.lean
```diff
@@ -43,6 +82,43 @@ nr_def «rotate_left»<>(u : u8, N : u8) -> u8 {
...
+    let x5 = #uAnd(#uAnd(x2, x3) : u8, x4) : u8;
+    let x6 = (@rotate_left<> as λ(u8, u8) → u8)(x5, 1 : u8);
+    #uXor(v, x6) : u8;
+}
+
+nr_def «sgn0»<>(self : Field) -> u1 {
+    #cast(self) : u1;
+}
+
+nr_def «mtree_recover»<H, @N : type_u 32>(idx : [bool; N], p : [Field; N], item : Field) -> Field {
```
kustosz, line 99: would it be very difficult to reuse the typelevel syntax here? Like `@N : u32` and `@N : Field`?

Lampe/Lampe/Syntax.lean
```diff
@@ -44,10 +44,12 @@ syntax "λ(" nr_type,* ")" "→" nr_type : nr_type -- Function
...
 syntax "@" nr_ident "<" nr_generic,* ">" : nr_type -- Type alias
 
-syntax num ":" num : nr_generic
+syntax num ":" "type_u" num : nr_generic
+syntax num ":" "type_fp" : nr_generic
 syntax nr_type : nr_generic
 
-syntax "@" ident ":" num : nr_generic_def -- Kind.u
+syntax "@" ident ":" "type_u" num : nr_generic_def -- Kind.u
+syntax "@" ident ":" "type_fp" : nr_generic_def -- Kind.field
```
kustosz, line 52: Actually would be easy, you just go `"@" ident ":" ident` here and then check if the ident is "whitelisted" in the elaborator. Let's do it!

kustosz (approved):
Awesome stuff, just a small request to make it somewhat nicer

mpenciak:
Numeric generic types have a much nicer syntax mirroring that of regular numbers now. Let me know how it looks!

kustosz (approved):
Approving once again to say this is awesome! :)

PR #79, opened 2025-03-26
Added generation based on noir project.

piohei:
This PRs introduces new way to generate output. Instead of single file whole Noir project is loaded and structure is being generated by the Lampe tool with for now only single `src/main.nr` extracted.


Examples/Dependencies/lampe/Dependencies/lakefile.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "Dependencies"
+version = "0.0.0"
+defaultTargets = ["Dependencies"]
+
+[[lean_lib]]
+name = "Dependencies"
+
+[[require]]
+name = "Lampe"
+path = "./../../../../Lampe"
```
kustosz, line 11: if this is autogenerated, it should point to `lampe` on github and not local

Examples/test.sh
```diff
@@ -13,17 +13,22 @@ readarray -t example_dirs < <(find "$EXAMPLES_DIR" -maxdepth 1 -type d -not -pat
...
-	mv $extracted $extracted_temp
+  lib_name=$(grep -A1 "^\[package\]" "Nargo.toml" | grep "name" | sed 's/name = "\(.*\)"/\1/')
 
-	$CLI --root $dir"/src" --out-file $extracted
+	if [[ "$lib_name" =~ Dependencies|LocalDependency ]]; then
+	  continue
+	fi
+
+	$CLI
+	if [[ "$lib_name" =~ Merkle ]]; then
```
kustosz, line 23: Why do we need this? If this gets overwritten by the CLI, there should be no user-generated stuff in there, so we need to restructure the project

kustosz (commented):
Cool stuff, asked a few questions inline, but this looks promising!

Examples/Dependencies/lampe/Dependencies/lakefile.toml
piohei: I thought that it is important for you to test for now. I've copied the comment and code with github commented is here https://github.com/reilabs/lampe/blob/piohei/noir_dependency_support/src/file_generator/lake/mod.rs#L20

Examples/test.sh
piohei: yeah - this is something I know we need to change as I agree 100%.  I'm not sure how yet - need to catch up on Lean ecosystem first.

piohei: I've change the generated structure in latest commits.

Examples/test.sh
```diff
@@ -4,26 +4,71 @@ set -euxo pipefail
...
 
-	$CLI --root $dir"/src" --out-file $extracted
+	cd lampe
 
 	lake exe cache get
 	lake build
 
-	rm $extracted
-	mv $extracted_temp $extracted
+	# We need to cleanup some space in GitHub as we run out of disk space when running tests
```
piohei, line 68: I needed to add this hack as we run out of disk space during e2e-test CI run.

Examples/MerkleFromScratch/user_actions.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/bash
```
piohei, line 1: Here is file simulating "user work" with the generated files.

kustosz (approved):

PR #81, opened 2025-03-29
Slight improvement to `sl` tactic

mpenciak:
This PR adds slightly more to the `sl` tactic to close goals of the form found in #77.

### DIFF
--- a/Lampe/Lampe/Tactic/SL.lean
+++ b/Lampe/Lampe/Tactic/SL.lean
@@ -224,6 +224,12 @@ lemma solve_compose [LawfulHeap α] {P Q R S : SLP α} (h₁ : P ⊢ Q ⋆ R) (h
   apply SLP.star_mono_l
   assumption
 
+theorem solve_pure_ent_pure [LawfulHeap α] {P Q : Prop} :
+  (P → Q) → ((⟦P⟧ : SLP α) ⊢ ⟦Q⟧) := by
+  unfold SLP.lift
+  unfold SLP.entails
+  simp only [and_imp, forall_eq_apply_imp_iff, and_true, imp_self]
+
 end Internal
 
 structure SLGoals where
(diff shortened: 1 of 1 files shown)


Lampe/Lampe/Tactic/SL.lean
```diff
@@ -347,6 +353,17 @@ partial def solveEntailment (goal : MVarId): TacticM SLGoals := withTraceNode `L
   trace[Lampe.SL] "Current goal: {pre.printShape} ⊢ {post.printShape}"
   trace[Lampe.SL] (←Lean.PrettyPrinter.ppExpr target)
 
+  match pre, post with
```
kustosz, line 356: I'd like this to be unified into the structure that we already have – adding a pure-pre to the huge match below and then solving this goal in a separate function, requiring the RHS to be pure as well.

Lampe/Lampe/Tactic/SL.lean
```diff
@@ -347,6 +353,17 @@ partial def solveEntailment (goal : MVarId): TacticM SLGoals := withTraceNode `L
...
+  | SLTerm.lift _, SLTerm.lift _ => do
+    let g :: impls ← goal.apply' (←mkConstWithFreshMVarLevels ``Internal.solve_pure_ent_pure) | throwError "unexpected goals in solve_pure_ent_pure"
+    let (_, g) ← g.intro1
+    let gs ← evalTacticAt (←`(tactic|tauto)) g
```
kustosz, line 360: Are you sure this works? I think `tauto` fails if it solves the goal, which will throw you out of `sl`, back to before `solve_pure_ent_pure` was called

kustosz (changes requested):

Lampe/Lampe/Tactic/SL.lean
```diff
@@ -319,6 +327,22 @@ partial def solvePureStarMV (goal : MVarId) (lhs : SLTerm): TacticM SLGoals := w
       pure $ SLGoals.mk [final] impl
   | _ => throwError "Unrecognized shape in solvePureStarMV"
 
+/--
+Solves goals of the form `⟦P⟧ ⊢ ⟦Q⟧` by transforming it to a goal of the form `P → Q` and trying
+-/
+def solvePureEntailMV (goal : MVarId) (pre : SLTerm) : TacticM SLGoals := withTraceNode `Lampe.SL (fun e => return f!"solvePureEntailMV {Lean.exceptEmoji e}") do
+  let g :: impls ← goal.apply' (←mkConstWithFreshMVarLevels ``Internal.solve_pure_ent_pure) | throwError "unexpected goals in solve_pure_ent_pure"
+  let gs ← evalTacticAt (←`(tactic|try tauto)) g
```
kustosz, line 335: once again, I think this is broken – `tauto` will throw if it cannot close the goal and then it throws you out of this block – you need a `try` here

Lampe/Lampe/Tactic/SL.lean
```diff
@@ -319,6 +327,22 @@ partial def solvePureStarMV (goal : MVarId) (lhs : SLTerm): TacticM SLGoals := w
...
+  if gs.isEmpty then
+    return SLGoals.mk [] impls
+  else
+    -- If we are left in a `True → P` goal state then replace this with `P`
+    if pre == .lift (mkConst ``True) then
```
kustosz, line 340: not sure if this is correct, there could be some universe level shenanigans that make it not-equal. `isAppOf` is probably a better choice

kustosz (changes requested):

Lampe/Lampe/Tactic/SL.lean
mpenciak: The evaluation is of ``` `(tactic|try tauto) ```

kustosz (approved):

PR #86, opened 2025-04-30
Add Github Action to run noir frontend tests with Lampe.

piohei:
GH Action may not work yet but I need to merge it to main to be able to test it. GH will not add action that is not defined on main branch.


mpenciak (approved):
Looks good!

PR #90, opened 2025-05-05
Support dependencies for project.

piohei:


mpenciak (approved):
Looks good to me! 

One last minor nit is that maybe the functions in `src/file_generator/lean/mod.rs` could have doc comments describing which function generates which file in the project, as the names are a little confusing and hard to match up with the file. 

We can either do this now or later, I have no preference.

Otherwise thanks for this! And thank you for your patience with the long back and forth we had on this PR.

piohei:
I will merge and add docs as next PR so all other PRs can use new structure already. :)

PR #93, opened 2025-05-13
fix: Matching generics in traits and structs

mpenciak:
This PR fixes a small issue where if the trait and type both use the same generic type variable in a trait impl block, then the extractor would extract the generic parameter multiple times.

An example is

```rust
trait Foo<I> {
    fn foo(self) -> I
}

struct Pair<I> {
    a : I,
    b : I,
}

impl<I> Foo<I> for Pair<I> {
    fn foo(self) -> I {
        self.a
    }
}
```

used to extract to

```lean
nr_struct_def multiple_generics::Pair<I> {
    a : I,
    b : I
}

nr_trait_impl[impl_405] <I, I> multiple_generics::Foo<I> for multiple_generics::Pair<I> where  {
    fn «multiple_generics»::«foo»<> (self : multiple_generics::Pair<I>) -> I {
        (self as multiple_generics::Pair<I>).a;
}
}
```

But now extracts to

```lean
nr_trait_impl[impl_405] <I> multiple_generics::Foo<I> for multiple_generics::Pair<I> where ...
```

This PR also tries to slightly clean up the end to end testing infrastructure. With the merging of #84 we can now lump a number of tests into a single noir/lampe project.

To this end I've added an `ExtractionTests` folder that starts the process of merging some of the simpler tests. For tests in the `Examples/` folder that only check whether Lampe successfully extracts and parses the resulting extracted Lean correctly, we can start adding the tests as separate files in the `ExtractionTests` folder.


iamrecursion (approved):
This looks good to me, with the caveat that I am not overly familiar with the current state of the extractor. Nevertheless, the logic and the extracted result both make intuitive sense to me.

PR #94, opened 2025-05-13
fix: namespaces for impls and structs

mpenciak:


src/lean/mod.rs
```diff
@@ -1191,21 +1196,31 @@ impl<'file_manager, 'parsed_files> LeanEmitter<'file_manager, 'parsed_files> {
...
                                     Some(self_type) => {
                                         let self_type_str =
                                             self.emit_fully_qualified_type(self_type, ctx);
-                                        format!("{self_type_str}::{name}")
+
+                                        // If the self-type is a slice or array, we treat this
+                                        // differently and use the `std::slice` or `std::array` as
+                                        // the namespace. This is what is detected by the builtin
+                                        // syntax emitter to detect whether something is a builtin
+                                        // or not.
```
iamrecursion, line 1208: Oh this is GROSS. I don't see how you can do better though until we address #102 though, so not a blocker to merging this imo.

iamrecursion (approved):
LGTM! A simple—if gross—fix for a reasonably simple issue!

PR #98, opened 2025-05-22
Add docs.

piohei:
Docs for code to generate files.
I could add flag like "debug file creation" that could add as comment to each generated file method used in rust to create it. Let me know what you think as in my opinion we probably won't be doing a lot of changes in that part of lampe so the time invested for it will probably won't give proper return on the investment. ;)


mpenciak:
Adding the "debug file creation" flag could be helpful... But really just adding something like "This generates the `./Package/Extracted/Package.lean` file... This generates the `./Package.lean` file... etc..." would be enough to help with file generation understandability.

piohei:
Great idea! I added some example paths to methods that create files. I hope it will help. :)

PR #125, opened 2025-06-02
Piohei/docs

piohei:


iamrecursion:
Regarding the "to consider" section, I don't think we care about installing Noir as that is technically not required for the project. A worked example of Merkle would be nice, but I think that is something for later. Please file an issue for doing this and schedule it for milestone MX.

piohei:
> Regarding the "to consider" section, I don't think we care about installing Noir as that is technically not required for the project. A worked example of Merkle would be nice, but I think that is something for later. Please file an issue for doing this and schedule it for milestone MX.

Issue added!

iamrecursion (approved):
Other than that LGTM!

PR #120, opened 2025-05-28
Delaborate `Tp.denote`

mpenciak:
This PR adds a delaborator for `Tp.denote`. This is part of a long effort to make the proof states
the user will encounter less "scary".

PR #132, opened 2025-06-04
fix: `Tp.denote` delaborator

mpenciak:
Fix the delaborator for `Tp.denote` to avoid any infinite recursion that was observed after #120 and #126. I do not foresee any issues from this change, and it appears it fixes any of the delab issues in #125


iamrecursion (approved):
Looks good to me. This fix is _much_ cleaner!

PR #134, opened 2025-06-04
Improve formatting

iamrecursion:
We were historically using stable `rustfmt` but this doesn't support a lot of formatting directives. This upgrades the repo to use nightly instead, to enable better formatting. Run it with `cargo +nightly fmt`, and it will be checked on CI.


piohei (approved):
Nice! I like it :D

Issue #139, opened 2025-06-06
Fix namespacing issues with trait extraction and resolution


iamrecursion:
Currently the combination of our extraction from Noir and the trait-resolution mechanism is a bit broken, preventing us from extracting more complex projects. This needs to be fixed.

mpenciak:
Currently, the following example behaves incorrectly:

The noir snippet
```rust
pub mod foo_struct {
    pub struct FooStruct {
        pub x: Field,
    }
}

pub mod bar_trait {
    pub trait BarTrait {
        fn bar(self) -> Field;
    }
}

mod baz_impl {
    use super::bar_trait::BarTrait;
    use super::foo_struct::FooStruct;

    impl BarTrait for FooStruct {
        fn bar(self) -> Field {
            self.x
        }
    }
}

mod use_impl {
    use super::foo_struct::FooStruct;
    use super::bar_trait::BarTrait;

    fn randomfunc(t: FooStruct) -> Field {
        t.bar()
    }
}
```

extracts to
```
nr_struct_def foo_struct::FooStruct<> {
    x : Field
}

...

nr_trait_impl[impl_428] <> bar_trait::BarTrait<> for foo_struct::FooStruct<> where  {
    fn «baz_impl»::«bar»<> (self : foo_struct::FooStruct<>) -> Field {
        (self as foo_struct::FooStruct<>).x;
}
}

nr_def «use_impl»::«randomfunc»<>(t : foo_struct::FooStruct<>) -> Field {
    ((foo_struct::FooStruct<> as BarTrait<>)::bar<> as λ(foo_struct::FooStruct<>) → Field)(t);
}

```

The issue is that the implemented trait function inherits the namespace in which it is defined. The current behavior of the `enter_trait` tactic relies on the extractd impl block to behave like:
```
...
fn «bar»<> (self : foo_struct::FooStruct<>) -> Field {
...
```

Issue #138, opened 2025-06-06
Merkle (skyscraper) into multifile (multimodule) project


piohei:
We want to make sure that the skyscraper (merkle) demo is a multifile project, and that it works. Tutorial for it is another issue created.

Issue #109, opened 2025-05-22
Prove monotonicity theorems for use in tactics


iamrecursion:
Currently we assume that we have a closed-world with our Lean proofs. We need to implement some theorems about monotonicity so that people can import a noir library and similarly import our theorems and use them to reason about things.

PR #143, opened 2025-06-10
Fix Lampe trait resolution tactic

mpenciak:
This PR fixes two issues (and closes issue #139) 

* The previous `enter_trait ...` tactic was not able to introspect on the trait implementations in compound environments of the form `def compoundEnv : Lampe.Env := env1 ++ env2`

* The `try_impls_all ...` tactic was too eager to unify some metavariables when used as a part of `enter_trait`. The check that the tactic was successful did not take into account whether the outside call to `callTrait_direct_intro` was able to match on the correct trait, and this lead to false successes for `try_impls_all`.

The solution is to replace the tactic `enter_trait ...` with a new `try_all_traits` tactic which combines the call to `callTrait_direct_intro` and `try_impls_all`. The tactic then unifies all of the resulting metavariables from the function applications in the right order for the type inference system to do the rest of the job.

With this PR, issue #138 is unblocked.

Note: We will likely need to revisit this tactic when we get around to #109, #144

### DIFF
--- a/Lampe/Lampe/Tactic/Traits.lean
+++ b/Lampe/Lampe/Tactic/Traits.lean
@@ -12,17 +12,17 @@ namespace Lampe
 
 theorem STHoare.callTrait_direct_intro {impls : List $ Lampe.Ident × Function}
     (h_trait : TraitResolution Γ ⟨⟨traitName, traitKinds, traitGenerics⟩, selfTp⟩ impls)
-    (h_fn : impls.find? (fun (n, _) => n = fnName) = some (fnName, func))
-    (hkc : func.generics = kinds)
-    (htci : (func.body _ (hkc ▸ generics) |>.argTps) = argTps)
-    (htco : (func.body _ (hkc ▸ generics) |>.outTp) = outTp)
-    (h_hoare: STHoare p Γ H (htco ▸ (func.body _ (hkc ▸ generics) |>.body (htci ▸ args))) Q) :
+    (h_find_fn : impls.find? (fun (n, _) => n = fnName) = some (fnName, func))
+    (h_kinds_eq : func.generics = kinds)
+    (h_args_eq : (func.body _ (h_kinds_eq ▸ generics) |>.argTps) = argTps)
+    (h_out_eq : (func.body _ (h_kinds_eq ▸ generics) |>.outTp) = outTp)
+    (h_pre_hoare: STHoare p Γ H (h_out_eq ▸ (func.body _ (h_kinds_eq ▸ generics) |>.body (h_args_eq ▸ args))) Q) :
     STHoare p Γ H (Expr.call argTps outTp (.trait (some selfTp) traitName traitKinds traitGenerics fnName kinds generics) args) Q := by
   apply STHoare.callTrait_intro (selfTp := selfTp) (traitName := traitName) (traitKinds := traitKinds) (traitGenerics := traitGenerics) (fnName := fnName) (outTp := outTp) (generics := generics)
   · simp [SLP.entails_top]
   · exact h_trait
-  · simp only [Option.eq_some_iff_get_eq] at h_fn
-    cases h_fn
+  · simp only [Option.eq_some_iff_get_eq] at h_find_fn
+    cases h_find_fn
     rename_i h
     rw [←h]
     simp [List.get_find?_mem]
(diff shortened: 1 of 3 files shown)


Lampe/Lampe/Tactic/Traits.lean
```diff
@@ -5,84 +5,158 @@ import Lampe.Syntax
...
+
+namespace Lampe
+
+theorem STHoare.callTrait_direct_intro {impls : List $ Lampe.Ident × Function}
+    (h_trait : TraitResolution Γ ⟨⟨traitName, traitKinds, traitGenerics⟩, selfTp⟩ impls)
+    (h_fn : impls.find? (fun (n, _) => n = fnName) = some (fnName, func))
+    (hkc : func.generics = kinds)
+    (htci : (func.body _ (hkc ▸ generics) |>.argTps) = argTps)
+    (htco : (func.body _ (hkc ▸ generics) |>.outTp) = outTp)
+    (h_hoare: STHoare p Γ H (htco ▸ (func.body _ (hkc ▸ generics) |>.body (htci ▸ args))) Q) :
```
iamrecursion, line 19: These names really are not great. It might be worth giving these more explanatory names, as it is pretty hard to work these out without really picking apart the definition of the theorem.

Lampe/Lampe/Tactic/Traits.lean
```diff
@@ -5,84 +5,158 @@ import Lampe.Syntax
...
+  · simp only [Option.eq_some_iff_get_eq] at h_fn
+    cases h_fn
+    rename_i h
+    rw [←h]
+    simp [List.get_find?_mem]
+  · assumption
+  · assumption
+  · assumption
+
+/-- Extract all the `TraitImpl`s from a `Lampe.Env` -/
```
iamrecursion, line 33: Does this have any performance gotchas or potential to overflow the stack?

Lampe/Lampe/Tactic/Traits.lean
```diff
@@ -5,84 +5,158 @@ import Lampe.Syntax
...
+  | head :: tail =>
+    return ← mkAppOptM `HList.cons
+      #[← mkAppM `Lampe.Kind #[], repExpr, listExprs.head!, ← mkListExpr listExprs.tail, head,
+        ← mkHListExpr repExpr tail listExprs.tail]
 
 /--
-`try_impls` takes a comma separated sequence of `TraitImpl`s in square brackets,
-and it tries them one by one by calling `apply_impl` until one of them succeeds.
+Tactic to make progress on goals of the form `Expr.call (FuncRef.trait ...)` by trying all of the
+trait implementations in a given environment
```
iamrecursion, line 86: I think this documentation needs to explain a bit more about the algorithm, and why it is done this way, as well as the potential performance pitfalls (if any) with large `Env`s, etc.

Lampe/Lampe/Tactic/Traits.lean
```diff
@@ -5,84 +5,158 @@ import Lampe.Syntax
...
+
+    for func in funcs do
+      let funcArgs := func.getAppArgs
+      let funcFunc := funcArgs.reverse[0]!
+      let funcName := funcArgs.reverse[1]!
+
+      trace[Lampe.Traits] m!"  attempting function {← ppExpr funcName}"
+      -- In order for the metavariables to unify in the right way, we need to run these
+      -- tactics in this order as otherwise `tauto` is overly eager to fill in data
+      -- incorrectly
```
iamrecursion, line 126: How was `tauto`s eagerness discovered here? Just trial and error? Please give _any_ useful information for someone who might have to modify this tactic in the future!

Lampe/Lampe/Tp.lean
```diff
@@ -107,6 +109,26 @@ def Kind.denote : Kind → Type
...
+| [], _ :: _
+| _ :: _, [] => isFalse (by simp only [List.nil_eq, reduceCtorEq, not_false_eq_true])
+| a :: as, b :: bs => match kindDecEq a b, kindsDecEq as bs with
+  | isTrue h, isTrue hs => isTrue (by simp only [h, hs])
+  | isFalse h, _ => isFalse (by simp [h])
+  | _, isFalse h => isFalse (by simp [h])
+
+instance : DecidableEq Kind := kindDecEq
+
+instance : DecidableEq $ List Kind := kindsDecEq
```
iamrecursion, line 130: For completeness we might as well add an `instance : DecidableEq $ List Tp := tpsDecEq` as that does not exist up above but might as well!

Lampe/Tests/Envs.lean
```diff
@@ -0,0 +1,144 @@
+import Lean
```
iamrecursion, line 1: This test should be cleaned up quite a bit, and minimized to exactly what is needed to demonstrate the issue should trait resolution regress at any point.

iamrecursion (changes requested):
This is good stuff, but I think there is still some cleaning up that is necessary before this lands! It shouldn't take too long though, I hope!

Lampe/Lampe/Tactic/Traits.lean
mpenciak: I don't believe so... The function just collects the impls from an environment. It'll go linearly through a cons list, but it splits on envs of the form `env1 ++ env2`, so the stack shouldn't be as deep as the whole list of traits.

mpenciak:
@iamrecursion Just fixed up some of the issues you pointed out, let me know how it looks!

iamrecursion (approved):
Looks great to me!

PR #151, opened 2025-06-18
Fix handling --root param.

piohei:
When relative path is being used for the project noir has some bug and relative path is joined twice. Example: `./testing/Merkle/teting/Merkle/src/main.nr`. Then compiler can't find proper files. Nargo is using absolute path so instead of fixing the bug we can follow their approach.


iamrecursion (approved):
LGTM!

PR #159, opened 2025-06-23
Added noir_stdlib copy with extraction script.

piohei:
This is just adding noir stdlib to our codebase with script to run extraction using lampe from same commit. As of now it is not working due to other missing bits. Checking noir versions consistency will be done in next PR.
I was thinking about using sparse git submodule for stdlib but we generate lampe output in noir's project. Then any code added in that directory would be trated as changes to submodule itself instead of lampe. I also don't want to allow to have lampe extracted in random place for the noir's project to keep it together.


stdlib/extract.xsh
```diff
@@ -0,0 +1,19 @@
...
+	script_dir = Path($(echo $XONSH_SOURCE).strip()).resolve()
+	return script_dir.parent.parent
+
+project_root = get_project_root()
+
+(cd @(project_root) && cargo build --release)
+
+lampe_cmd = project_root / 'target' / 'release' / 'lampe'
+
+cd @(project_root / 'noir_stdlib')
```
iamrecursion, line 17: This is still wrong, no?

iamrecursion (approved):
Looks good other than the issue in the extraction script.

PR #161, opened 2025-06-24
Prove environment monotonicity for `STHoare`

### LINKED ISSUE #109: Prove monotonicity theorems for use in tactics

iamrecursion:
In order to compose Noir libraries with their accompanying lampe proofs, it is necessary to be able to extend proofs performed over some environment e_1 to any environment e_2 that contains e_1. In other words, we need to prove that our language semantics are _monotonic_ with respect to their environment.

To this end, this commit states and proves a theorem for the monotonicity of `Omni` and then `STHoare` using that. It also contains proofs of the monotonicity of `TraitResolvable` and `TraitResolution` as these proved necessary. It also includes a subset relation on `Env`s, and provides a number of useful lemmas for reasoning about this subset relation.

Furthermore, it adds a test that demonstrates the correct function of the monotonicity theorems. This test performs a number of proofs of functions in isolation, both using traits and not, and shows that those proofs can be used in proofs that take place in an environment of which each proof's environment was a subset.

Closes #109.


Lampe/Lampe/Hoare/SepTotal.lean
```diff
@@ -1,27 +1,35 @@
...
-if `P` holds, then `e` terminates and it either (1) fails or (2) *successfully* evaluates to `v` such that `Q v` holds.
-This is logically equivalent to saying that if `P` holds, then `e` terminates and (`Q v` holds if `e` succeeds and evaluates to `v`).
-
-Hence, the triples are *partial* with respect to failure and *total* with respect to termination.
-
-An intuitive way of looking at this is thinking in terms of "knowledge discovery".
-For example, if the operation `a + b` succeeds, then we know that it evaluates to `v = a + b` **and** `a + b < 2^32`, i.e., no overflow has happened.
-Then, we would define the post-condition such that `Q = λv ↦ (v = a + b) ∧ (a + b < 2^32)`.
- -/
+-- A Hoare triple where the states must be valid under separation logic.
```
kustosz, line 9: Why this change? It goes against mathlib's style for docstrings (https://leanprover-community.github.io/contribute/doc.html)

Lampe/Lampe/Semantics/Env.lean
```diff
@@ -0,0 +1,116 @@
...
+
+-- The following namespace contains lemmas that are necessary to reason about the composition of
+-- our environments when using proofs across multiple libraries.
+--
+-- Note that they are all tagged with `@[simp]` to be made available to the simplifier for ease of
+-- use.
+namespace Env
+
+@[simp]
+lemma append_functions {Γₗ Γᵣ : Env} : (Γₗ ++ Γᵣ).functions = Γₗ.functions ++ Γᵣ.functions := by
```
kustosz, line 44: This should be called `functions_append`

Lampe/Lampe/Semantics/Env.lean
```diff
@@ -0,0 +1,116 @@
...
+  rfl
+
+@[simp]
+lemma append_traits {Γₗ Γᵣ : Env} : (Γₗ ++ Γᵣ).traits = Γₗ.traits ++ Γᵣ.traits := by
```
kustosz, line 48: same comment as above

Lampe/Lampe/Semantics/Env.lean
```diff
@@ -0,0 +1,116 @@
...
+  rfl
+
+@[trans]
+lemma subset_transitive {Γ₁ Γ₂ Γ₃ : Env} : Γ₁ ⊆ Γ₂ → Γ₂ ⊆ Γ₃ → Γ₁ ⊆ Γ₃ := by
```
kustosz, line 52: there's a convention to shorten `transitive` to `trans`

kustosz (approved):
Pre-approving, but there's a bunch of (uninteresting) stuff to fix

Lampe/Lampe/Hoare/SepTotal.lean
```diff
@@ -1,27 +1,37 @@
...
-Then, we would define the post-condition such that `Q = λv ↦ (v = a + b) ∧ (a + b < 2^32)`.
- -/
+A reasonable intuition for how this works is the lens of "knowledge discovery". For example, if the
+operation `a + b` on `u32` succeeds, then we know that it evaluates to `v = a + b` **and** that
+`a + b < 2^32` (that no overflow has happened). Under such circumstances, we would define the
+postcondition such that `Q = λv ↦ (v = a + b) ∧ (a + b < 2^32)`.
+
+Note that we have the `⋆ ⊤` term in our postcondition here. This term is able to subsume evidence
+that goes unused, thereby turning the separation logic from linear to affine. This is necessary to
+model programs that have unused variables and other such foibles of real world code.
```
kustosz, line 33: This is necessary for a deeper reason, not really "weird programs". It let's you drop assumptions that you don't need in perfectly normal code. For example if you consider something like
```
fn sum(slice: [Field]) -> Field {
    let mut result = 0;
    for a in slice {
        result += a;
    }
    result
}
```
the triple `[[]] sum(x) (fun v => v = lean_sum x)` is not provable linearly - you'd have to track the existence of that internal mutable variable in the postcondition _forever_. Linear logic is workable only in languages that have explicit memory freeing. Affine logic let's you model garbage collected languages, where instead of the explicit `drop` expression, you prune the postconditions based on vibes.

kustosz (approved):

PR #164, opened 2025-06-25
Trait definitions

### LINKED ISSUE #160: Improve trait expressions for proofs
While we have `func_name.call` as shorthand for calling a noir function in a program expression as part of an `STHoare` triple, there is no such equivalent for functions in trait implementations.

Something like the following is currently necessary:

```lean
STHoare 
  lp 
  env
  ⟦True⟧
  (Expr.call [Tp.field, Tp.field] Tp.field
    (FuncRef.trait H "hasher::BinaryHasher" [Kind.type] (HList.cons Tp.field HList.nil) "hash" [] HList.nil) h![a,b])
  (fun v => v = H' (a ::ᵥ b ::ᵥ .nil))):
```

This is a bit of a monstrosity. Instead, we want to be able to write something more akin to the function case, such as `TraitName.methodName selfType h![trait_generics] h![method_generics] h![params]`.

kustosz:
Outputs utility defs for Noir trait defs, so that we can call them painlessly.

Closes #160.

### DIFF
--- a/src/lean/mod.rs
+++ b/src/lean/mod.rs
@@ -628,7 +628,12 @@ impl<'file_manager, 'parsed_files> LeanEmitter<'file_manager, 'parsed_files> {
             })
             .join("\n");
         indenter.dedent()?;
-        Ok(syntax::format_trait_def(&name, &generics, &associated_types, methods))
+        Ok(syntax::format_trait_def(
+            &name,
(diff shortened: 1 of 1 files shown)


iamrecursion (approved):
This is such a nice usability enhancement! :D

PR #166, opened 2025-06-25
Migrate test scripts to xonsh

### LINKED ISSUE #135: Rewrite testing/test.sh to xonsh
It has started to be difficult to maintain the current testing scripts in bash. We have agreed to rewrite them all to [xonsh](https://xon.sh).

This issue includes adding a new option to overwrite generated files. It is required to support changes to the generation that occur during development.

mpenciak:
This PR migrates all of our test scripts from shell scripts to xonsh scripts to improve their maintainability and reliability. This closes #135.

It also adds functionality to the noir frontend test script:
* The `--use-local` flag can be given to use the local version of Lampe instead of the main branch on github
* The `--test` flag was made easier to use. You can invoke it with an absolute path to the test, or just something like `noir_test_success/should_fail_with_matches`.


testing/test.xsh
```diff
@@ -0,0 +1,147 @@
...
+        example_dirs = [examples_dir / selected_test]
+
+    for dir_path in example_dirs:
+        cd @(dir_path)
+        dir_name = dir_path.name
+
+        if dir_name.startswith('_'):
+            continue
+
+        clean_script = dir_path / "clean.sh"
```
piohei, line 61: Open question - should we also migrate `clean.sh`/`user_actions.sh` to xonsh?

testing/test.xsh
```diff
@@ -0,0 +1,147 @@
...
+        if clean_script.exists():
+            /usr/bin/env bash @(clean_script)
+
+        # Get the extracted dir (if it exists)
+        extracted_dirs = list(dir_path.rglob("Extracted"))
```
piohei, line 66: Shouldn't this be a single directory? Why list here?

testing/test.xsh
```diff
@@ -0,0 +1,147 @@
...
+                    if not checkedout_file.exists():
+                        print(f"NEW FILE: {generated_file} is newly generated")
+                        sys.exit(1)
+
+                # delete renamed files
+                for checkedout_file in extracted_dir.rglob("*.lean_checkedout"):
+                    checkedout_file.unlink()
+
+        # Overwrite Lampe to local path
+        lakefile_path = dir_path / "lampe" / "lakefile.toml"
```
piohei, line 113: As we have Python now I would rather do it in more elegant way. :) In bash there was no other way. The goal here is just to change Lampe dependency to be local one.

testing/test.xsh
```diff
@@ -0,0 +1,147 @@
...
+            cd @(lampe_dir)
+
+            lake_symlink = lampe_dir / ".lake"
+            if not lake_symlink.exists():
+                ln -s @(lake_dir) .lake
+
+            lake exe cache get
+            lake build
+
+if __name__ == "__main__":
```
piohei, line 146: I'm not sure if this is required in xonsh script. I guess single `main()` without if should be enough.

testing/test.xsh
mpenciak: I think we should! I can migrate all the shell scripts to xonsh as part of this PR

testing/test.xsh
mpenciak: So this is a bit tricky... It is a single directory for almost every example, except not the `TestDeps` example where there are a bunch of `Extracted` folders in all the dependencies (but no `Extracted` in the main project!). Do we want to filter that one out and treat it as having no `Extracted`?

testing/test.xsh
mpenciak: Update: `clean.sh` is so simple I'm leavin git as is, but I moved all the `user_actions.sh` to xonsh

iamrecursion (changes requested):
This is fundamentally good, but I would really like to be able to factor out common functionality as much as we can. I worry about the amount of duplication here.

Issue #92, opened 2025-05-09
Improve error handling in our tactics


piohei:
Code:

```
theorem t1 {lp}: STHoare lp env (⟦⟧)
    (TestDep.«0.0.0».Extracted.GitDep.«0.0.0».hello.call h![] h![])
      fun output => (String.mk output.toList) = hello2 := by
      -- fun output => output = (hello2, by rfl) := by
  enter_decl
  simp only [TestDep.«0.0.0».Extracted.GitDep.«0.0.0».hello]
  steps []
  subst_vars
  unfold hello2
  rfl
```

Notice that `env` here is implicitly defined and is empty env.

Lean output when function not found
```
tactic 'rfl' failed, the left-hand side
  List.find?
    (fun x =>
      match x with
      | { name := n, «fn» := «fn» } => decide (n = Extracted.GitDep.«0.0.0».hello.name))
    env.functions
is not definitionally equal to the right-hand side
  some { name := Extracted.GitDep.«0.0.0».hello.name, «fn» := ?m.876 }
env : Env
lp : Lampe.Prime
⊢ List.find?
    (fun x =>
      match x with
      | { name := n, «fn» := «fn» } => decide (n = Extracted.GitDep.«0.0.0».hello.name))
    env.functions =
  some { name := Extracted.GitDep.«0.0.0».hello.name, «fn» := ?m.876 }
```

iamrecursion:
This is actually a broader usability issue with many of the Lampe tactics. The one above is _one_ of the ones that fails quietly, but there are more. All of these should be updated to provide accurate failure reporting.

PR #175, opened 2025-07-28
add some error handling to steps tactic

mpenciak:
This PR improves the error handling for some of our tactics, addressing #92.

* `steps` flags unkonwn supplied lemmas instead of failing silently
* `enter_block_as` and `loop_inv` flag incorrectly formatted/nonsense pre and post conditions instead of failing silently
* `enter_decl` gives a more insightful error when the declaration is not in the environment 

It also adds configuration options to `steps` with the syntax
`steps <limit> [<lemmas>] +config1 +config2 -not_config3 ...`
So far only `+strict` is detected, but it leaves open the option for more fine-grained steps behavior

TODOs:
- [x] Improve `enter_decl` error handling
- [x] Figure out the issue with type unification in pre and post conditions
- [x] Figure out if we can get rid of the `term` field in `AddLemma`

### DIFF
--- a/Lampe/Lampe/Tactic/Steps.lean
+++ b/Lampe/Lampe/Tactic/Steps.lean
@@ -376,13 +376,50 @@ partial def stepsLoop (goals : TripleGoals) (addLemmas : List AddLemma) (limit :
     | none => return goals
   | none => return goals
 
+structure StepsConfig where
+  limit : Nat
+  addLemmas : List AddLemma
+  strict : Bool
+
 /--
 Takes a sequence of at most `limit` steps to attempt to advance the proof state by continually
 simplifying the goal.
 -/
-partial def steps (mvar : MVarId) (config : StepsConfig) (limit : Nat) (addLemmas : List AddLemma) : TacticM (List MVarId) := do
+partial def steps (mvar : MVarId) (config : StepsConfig)  : TacticM (List MVarId) := do
   let goals ← stepsLoop (TripleGoals.mk mvar [] []) config.addLemmas config.limit config.strict
   return goals.flatten
+
+
+lemma STHoare.pluck_pures : (P → STHoare lp Γ H e Q) → (STHoare lp Γ (P ⋆ H) e (fun v => P ⋆ Q v)) := by
+  intro h
+  simp_all [STHoare, THoare, SLP.pure_star_iff_and]
+
+theorem callDecl_direct_intro {p} {Γ : Env} {func} {args} {Q H}
+    (h_found : (Γ.functions.find? (fun ⟨n, _⟩ => n = fnName)) = some ⟨fnName, func⟩)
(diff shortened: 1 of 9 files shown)


Lampe/Lampe/Tactic/Steps.lean
```diff
@@ -198,7 +199,7 @@ structure AddLemma where
 def tryApplySyntaxes (goal : MVarId) (lemmas : List AddLemma): TacticM (List MVarId) := match lemmas with
 | [] => throwError "no lemmas left"
 | n::ns => do
-  trace[Lampe.STHoare.Helpers] "trying {n.term} with generalizeEnv: {n.generalizeEnv}"
+  trace[Lampe.STHoare.Helpers] "trying {← ppExpr n.expr} with generalizeEnv: {n.generalizeEnv}"
```
iamrecursion, line 202:
```suggestion
  trace[Lampe.STHoare.Helpers] "trying {←ppExpr n.expr} with generalizeEnv: {n.generalizeEnv}"
```

Lampe/Tests/Negatives.lean
```diff
@@ -17,11 +17,10 @@ example : STHoare p env ⟦⟧ (zero_by_another_name.fn.body _ h![] |>.body h![]
   simp only [zero_by_another_name]
   steps
 
-  step_as (⟦⟧) (fun v => v = -1)
+  step_as (⟦⟧) (fun v => v = (-1 : Fp p))
```
iamrecursion, line 20: What changed that this is annotation is now necessary?

Lampe/Tests/Syntax.lean
```diff
@@ -284,7 +284,7 @@ example : STHoare p ⟨[add_two_fields], []⟩ ⟦⟧
     fun (v : Tp.denote p .field) => v = 3 := by
   simp only [call_decl]
   steps
-  step_as (⟦⟧) (fun v => v = 3)
+  step_as (⟦⟧) (fun (v : Fp p) => v = 3)
```
iamrecursion, line 287: Same with this?

iamrecursion (changes requested):
Fundamentally this looks good, but there are a few changes that seem like a regression to me.

mpenciak:
So to explain why the annotations are necessary:

Before the pre and post conditions for all of our tactics were just pure syntax holes you could fill in that would later be used by the tactics. The main change this PR institutes is it elaborates these syntax terms and gives the user some feedback on whether the syntax is mal-formed, not type correct, or any other issues it may have.

The upside of this is we get actual errors when you give tactics nonsense (like bad loop invariants, post-conditions that don't make sense, additional lemmas to steps that don't exist, etc...). The downside is that now when you write down pre/post conditions you need to give terms that typecheck, and in these cases this means the metavariable types need to be known.

PR #177, opened 2025-08-26
Re-write the Lampe DSL

### LINKED ISSUE #43: Debug Rust stack trace size
Currently the CI (and my local machine) require the additional `RUST_MIN_STACK` environment variable to be set to 16MB for the tests to pass.

We should explore why this is happening, and remove this needed stack size bump.

### LINKED ISSUE #60: Extractor indentation
Extractor indents the lines incorrectly.  This needs to be fixed.

### LINKED ISSUE #101: Noir standard library extraction
In order to be able to prove properties of the standard library (#50), we need the ability to actually _extract_ said library. To that end, we need to extract it.

- We likely need a copy of the Noir stdlib that is available and extracted to _all_ Lampe projects.
- To that end, it makes sense to have it as a new build artefact in this project such that it can be exposed to all Lampe projects (e.g. `import Lampe.NoirStd` or similar).
- Doing this is going to require a way to figure out how we keep this up to date with the version of Noir we depend on; so perhaps a check on CI?

There is a lot to be thought about for how we can do this.

iamrecursion:
This commit is an absolute monster, containing the following work and features:

- An adaptation of the extractor to split the analysis phase from the DSL syntax generation phase, enabling much cleaner logic and easier modification. It ensures that all items are extracted in a consistent order so as not to break proofs when re-extracting.
- Re-working of the Lean-side syntax definitions and compiler in order to handle compilation in a more consistent way and without extraneous structure that resulted in unnecessary work in proofs.
- Ensures that blocks are properly handled in the syntax as expressions, resulting in a proper representation of the Noir program semantics after extraction and compilation.
- Adds additional tactics to the Lampe codebase to enable easier work within complex proofs. Similarly, some tactics have been renamed to ensure consistency, and they have had documentation added.
- Extracted the Noir standard library and ensures that the extracted codebase compiles without errors.
- Updated all existing DSL source to use the new syntax definition. The old syntax definition is deprecated and scheduled to be removed in the future.

Closes #101 
Closes #60
Closes #43


mpenciak:
lgtm

PR #188, opened 2025-08-27
Fix conflicts with Lean keywords

mpenciak:


iamrecursion (approved):
This looks like a nice upgrade!

PR #195, opened 2025-08-28
Ensure we gather methods on _all_ types

iamrecursion:
Previously we were using the fact that `impl` blocks for structs were declared as child modules to gather methods, but this did not work for impls on builtin types.

This commit changes how we do method discovery for a file in three ways:

- When a module `is_type` (was created to hold impl methods) we ignore it at the top level.
- When traversing the entries in a module, we now identify type definitions and extract their direct methods, generating them if they are defined in the same file.
- When extracting the standard library, we additionally query for methods on builtin types, and add them to the current file if they were defined in that file.

Additionally, for methods or functions that directly represent calls to builtins, we no longer generate a stub method and instead generate a call to the builtin directly where one is encountered.


src/lean/generator.rs
```diff
@@ -704,13 +710,28 @@ impl LeanGenerator<'_, '_> {
...
+        // let file_name = self
+        //     .context
+        //     .file_manager
+        //     .path(id)
+        //     .unwrap()
+        //     .to_str()
+        //     .unwrap()
+        //     .to_string();
+        //
+        // if file_name.contains("array/mod.nr") {
```
kustosz, line 722: leftover

PR #202, opened 2025-09-02
Fix operator-trait naming

iamrecursion:
This commit ensures that, when an operator resolves to a trait implementation instead of a builtin, that we fully qualify the name of the trait in question. Previously they were being output unqualified, which was leading to name resolution issues.


mpenciak (approved):
One small question about keyword quoting from #188, but otherwise looks good.

Issue #50, opened 2025-01-23
Noir standard library formalization


mpenciak:
This is a tracking issue for the Noir standard library formalization.

PR #211, opened 2025-09-16
Add the remaining theorems for `Option`

iamrecursion:
This commit adds statements and proofs of the remaining theorems to accompany the functionality of the `std::option::Option<T>` type. It also introduces a document that aims to collate useful tips for stating theorems and proving.

Do note that it does not contain proofs for some of the traits implemented for `Option` as these will be influenced by the "lawful" trait initiative, and hence are not worth doing at this stage.

It also adds the `unsafe` mode to both the `steps` and `sl` tactics. This allows `sl` (and uses thereof embedded in `steps`) to unify from the entire proof state instead of the state of the current goal.

This is work toward #50.

### DIFF
--- a/Lampe/Lampe/Tactic/SL.lean
+++ b/Lampe/Lampe/Tactic/SL.lean
@@ -150,10 +150,23 @@ theorem move_to_sinks {p} {P Q : SLP (State p)} : (P ⊢ Q) → (P ⊢ (⟦⟧ �
 
 end Internal
 
+structure SLConfig where
+  isUnsafe : Bool
+
+def SLConfig.default : SLConfig := { isUnsafe := false }
+
+instance : Inhabited SLConfig where
+  default := SLConfig.default
+
 structure SLGoals where
   entailments : List MVarId
   props : List MVarId
   implicits : List MVarId
+
+abbrev SLM := ReaderT SLConfig TacticM 
+
+def SLM.run {α} (x : SLM α) (r : SLConfig) : TacticM α :=
+  ReaderT.run x r
 
 def SLGoals.flatten (g : SLGoals) : List MVarId := g.entailments ++ g.props ++ g.implicits
 
@@ -174,7 +187,7 @@ def _root_.Lean.MVarId.apply' (m: MVarId) (e: Lean.Expr): TacticM (List MVarId) 
 /--
 Solves goals of the form `P ⊢ [r ↦ v] ⋆ ?_`, trying to copy as much evidence as possible to the MVar on the right
 -/
-partial def solveSingletonStarMV (goal : MVarId) (lhs : SLTerm) (rhs : Lean.Expr) : TacticM SLGoals := 
+partial def solveSingletonStarMV (goal : MVarId) (lhs : SLTerm) (rhs : Lean.Expr) : SLM SLGoals := 
   goal.withContext $ withTraceNode `Lampe.SL (fun e => return f!"solveSingletonStarMV {Lean.exceptEmoji e}") $ do
   match lhs with
   | SLTerm.singleton _ v _ =>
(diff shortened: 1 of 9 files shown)


Lampe/Lampe/Tactic/SL.lean
```diff
@@ -226,18 +239,27 @@ For example, if the original goal is `P x ⋆ P y ⊢ (∃∃z, P z) ⋆ P x`, a
...
+      withNewMCtxDepth $ isDefEq expr rhs
+    if shouldProcessFinalGoal then
       let impl ← goal.apply' (←mkConstWithFreshMVarLevels ``Internal.entails_self_true)
       pure $ SLGoals.mk [] [] impl
     else throwError "final unrecognized is not equal"
   | SLTerm.star _ l r => do
     match l with
-    | SLTerm.unrecognized expr =>
-      if (←withNewMCtxDepth $ isDefEq expr rhs) then
+    | SLTerm.unrecognized expr => do
```
kustosz, line 257: you could write this as:
```
let mvar_guard <- if isUnsafe then id else withNewMCtxDepth
mvar_guard $ isDefEq expr this
```
Also `shouldProcessGoal` is a bad name for this variable. More like `shouldUnifyInGoal`?

Lampe/Lampe/Tactic/SL.lean
```diff
@@ -471,10 +493,23 @@ partial def solveEntailment (goal : MVarId): TacticM SLGoals :=
...
-elab "sl" : tactic => do
+/--
+Solves separation logic entailments of the form `P ⊢ Q ⋆ ⊤` or `P ⊢ Q ⋆ ?M`.
+
+It pushes all clonable information into the metavariable `?M` in order to strengthen it for usage in
+discharging further goals.
+
+It can be called in two ways:
+
+- `sl`: Simply solves the appropriate SL entailments, only unifying within the current goal context.
```
kustosz, line 504: This sentence makes no sense to me at all. Also this is more user facing, so to make it future-proof, I'd just say: `sl` is a safe tactic – it will make whatever progress it can while assuring it doesn't turn a solvable goal into an unsolvable one. `sl unsafe` is more aggressive, so it may close some more goals, but it may also get you stuck.

Lampe/Lampe/Tactic/Steps.lean
```diff
@@ -383,7 +383,7 @@ partial def step (mvar : MVarId) (addLemmas : List AddLemma) : TacticM TripleGoa
     let (_, hNext) ← hNext.intro (vname.getD `v)
     let hHead :: hEnt :: impls₂ ← hHead.apply (←mkConstWithFreshMVarLevels ``consequence_frame_left) | throwError "bad application"
     let impls₃ ← closer hHead
-    let rEnt ← solveEntailment hEnt
+    let rEnt ← SLMonad.run (solveEntailment hEnt) SLConfig.default
```
kustosz, line 386: I think you should give it a `.run`-free entry point, `SLM` shouldn't leak outside of that file.

Lampe/Lampe/Tactic/Steps.lean
```diff
@@ -516,14 +520,17 @@ It can be called in three main ways:
 - `steps`: A bare call to the tactic will simply try to advance the goal by performing as many steps
   as it can without blowing up the proof state or failing. This version can take a maximum of 10000
   steps before failing.
+- `steps unsafe`: As above, except that separation logic goals will be able to unify outside of the
```
kustosz, line 523: Again, I don't even know what "unifying outside of the current goal's context" means.

stdlib/lampe/Stdlib/Option.lean
```diff
@@ -30,91 +30,465 @@ export «std-1.0.0-beta.11».Extracted (
...
+@[simp]
+lemma option_snd_eq_toOption_get_of_isSome {p T}
+    {v : Tp.denote p $ «std::option::Option».tp h![T]}
+    (v_is_some : (toOption v).isSome = true)
+  : Builtin.indexTpl v Builtin.Member.head.tail = (toOption v).get v_is_some := by
+  rcases v with ⟨_|_, _, _⟩ 
+  . contradiction
+  . rfl
+
+lemma f_of_get_some_eq_map_f_of_some_getD {α β} {f : α → β} {d : β} {o : Option α}
```
kustosz, line 75: I would reverse the direction of this and call it `Option.getD_map_of_isSome`

stdlib/lampe/Stdlib/Option.lean
```diff
@@ -30,91 +30,465 @@ export «std-1.0.0-beta.11».Extracted (
...
+  . rfl
+
+lemma f_of_get_some_eq_map_f_of_some_getD {α β} {f : α → β} {d : β} {o : Option α} 
+    (h : o.isSome)
+  : f (o.get h) = (o.map f).getD d := by
+  rcases (Option.isSome_iff_exists.mp h) with ⟨_, rfl⟩
+  rfl
+
+@[simp]
+lemma f_of_bind_some_eq_f {α β} {v : Option α} {d : Option β} (v_is_some : v.isSome = true)
```
kustosz, line 82: This is a weird one, too much happening. I'd call this `Option.bind_def_of_isSome` and make it say `(v >>= f) = f (v.get h)`. This is definitionally equal in the case of a constant function, but much less clunky.

stdlib/lampe/Stdlib/Option.lean
```diff
@@ -30,91 +30,465 @@ export «std-1.0.0-beta.11».Extracted (
...
+    subst_vars
+    rw [f_of_get_some_eq_map_f_of_some_getD (f := func_emb) h]
+  . have h : (toOption v).isNone := by 
+      simp_all
+    steps [map_or_else_none_spec (E1 := E1) (E2 := E2) h default_f]
+    subst_vars
     simp_all
+
+theorem and_spec {p T v w P}
+  : STHoare p env P
```
kustosz, line 271: The `P` here is frame-able, so this should just say `STHoare p env [[]] ... (fun r => r = ...)`

stdlib/lampe/Stdlib/Option.lean
```diff
@@ -30,91 +30,465 @@ export «std-1.0.0-beta.11».Extracted (
...
+    (fun r => r = ((toOption self).map f_emb).getD (fromOption none) ⋆ P) := by
+  by_cases h : (toOption self).isSome = true
+  . steps [and_then_some_spec (E := E) h (f_f _)]
+    subst_vars
+    apply f_of_get_some_eq_map_f_of_some_getD h
+  . have h : (toOption self).isNone := by simp_all
+    steps [and_then_none_spec (E := E) h]
+    simp_all
+
+theorem or_spec {p T P self default}
```
kustosz, line 328: Another pointless `P`

stdlib/lampe/Stdlib/Option.lean
```diff
@@ -30,91 +30,465 @@ export «std-1.0.0-beta.11».Extracted (
...
+  by_cases h : (toOption self).isSome = true
+  . steps [or_else_some_spec h]
+    subst_vars
+    simp_all
+  . have h : (toOption self).isNone := by simp_all
+    steps [or_else_none_spec (E := E) h default_f]
+    subst_vars
+    simp_all
+
+theorem xor_spec {p T P self other}
```
kustosz, line 390: Pointless `P`.

Also jeeez what is this function even lol.

kustosz (approved):
All in all good stuff, some nitpicks inline

PR #217, opened 2025-10-01
Add overwrite flag to executable and fix its implementation

mpenciak:
This PR adds the `--overwrite` flag to the Lampe executable to control whether the Lampe-generated files get overwritten when run.

Before the behavior was inconsistent between the root package and its dependencies, with no way to control it.

Further we change the behavior of the overwrite logic to trigger by default on files with the `-- Generated by Lampe` comment at the top, with the flag overwriting this default behavior.


src/bin/main.rs
```diff
@@ -34,6 +34,10 @@ pub struct ProgramOptions {
     /// Testing mode?
     #[arg(long)]
     pub test_mode: bool,
+
+    /// Overwrite existing Lampe project if it exists.
```
iamrecursion, line 38:
```suggestion
    /// Overwrite all files if they exist, even if they were not generated by Lampe. Note that generated files are able to be overwritten by default.
```

Please wrap my suggestion properly!

iamrecursion (approved):
LGTM! Please just apply the one suggestion I've left!

PR #228, opened 2025-10-28
Clean up unused code in `EmitContext`

GarmashAlex:
Remove unused INCREASE_INDENT_TAIL, DECREASE_INDENT_TAIL, and split_up_funcref_name from 
src/lean/emit/context.rs.
These are obsolete: indentation is handled explicitly (no tail-inspection), and quoting is performed by the generator with environment-level whole-name quoting.


iamrecursion (approved):
Thanks for the contribution! I've let CI run on it.

PR #239, opened 2026-01-06
Add CI caching and improve CI tooling

Eduardogbg:
- fix our rust toolchain so it's not divergent between local and ci anymore; previously the local toolchain would be 1.90 but ci would use nightly features to format the code. to preserve the formatting options i opted into picking a nightly version close to 1.90 for formatting on ci and also overall local version, and 1.90 for other ci jobs. 
- install lake dependencies so we can use `lake exe cache get`
- use `lake exe cache get` to avoid rebuilding mathlib, lampe and stdlib
- fix out of disk issues on ci 
- create shared container image for all jobs, with dev dependencies already installed
  - xonsh and python deps
  - lake and c++ deps
  - cargo

### DIFF
--- a/scripts/test.xsh
+++ b/scripts/test.xsh
@@ -2,7 +2,10 @@ $RAISE_SUBPROC_ERROR = True
 
 import argparse
 from pathlib import Path
-import sys
+import os
+import re
+import shutil
+import subprocess
 import tempfile
 
 # --- Start of copied part.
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -11,38 +11,57 @@ jobs:
   ensure-noir-version:
     name: "Check Noir Versions"
     runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/eduardogbg/lampe-ci:latest
+      credentials:
+        username: Eduardogbg
+        password: ${{ secrets.GHCR_TOKEN }}
```
iamrecursion: Same comments as above with the namespace for these containers.

scripts/lean_cc_wrapper.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+
+
+def main() -> None:
```
iamrecursion: Why do we need this wrapper? Why do we care about not allowing someone using it to pass a specific linking type?

scripts/test.xsh
```diff
@@ -69,9 +89,47 @@ def run_tests(dir):
...
+
+def assert_generated_extracts(lampe_dir):
+    extracted_dirs = list(lampe_dir.glob("**/Extracted"))
+    extracted_files = list(lampe_dir.rglob("Extracted.lean"))
+
+    for extracted_dir in extracted_dirs:
+        for lean_file in extracted_dir.rglob("*.lean"):
+            contents = lean_file.read_text()
+            if LAMPE_GENERATED_COMMENT not in contents:
+                raise Exception(f"Non-generated file in extracted directory: {lean_file}")
```
iamrecursion: This is _technically_ not an error. Users are within their rights to put non-generated files in the generated directories, even if that seems stupid. The only thing we care about is that the files that were _actually extracted_ have the generated-by-lampe comment.

To do this kind of check properly you effectively want to be looking for every `*.lean` file that corresponds to an input `*.nr` file.

.github/workflows/ci-noir.yaml
```diff
@@ -8,59 +8,31 @@ jobs:
   run-tests:
     name: "Test Noir"
     runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/eduardogbg/lampe-ci:latest
+      credentials:
+        username: Eduardogbg
+        password: ${{ secrets.GHCR_TOKEN }}
```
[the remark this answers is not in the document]
Eduardogbg: Yes I know ahaha I just did it on my account so I could test everything else. I just asked Marcin to run a command:
```bash
export REILABS_GHCR_TOKEN='...'
echo "$REILABS_GHCR_TOKEN" | docker login ghcr.io -u <your-github-username> --password-stdin

docker buildx build --platform linux/amd64 \
    -t ghcr.io/reilabs/lampe-ci:latest \
    --push -f docker/ci/Dockerfile .
 ```
but you could run instead too. `REILABS_GHCR_TOKEN` just has to be a legacy PAT token with `write:packages` permission. Run it on repository root on this branch.

About the tags, sure it could be a SHA or maybe some other more easy to remember non-updatable tag. Then we could just hard-code it on the jobs yaml and at any PR update the image version we are using?

File: .github/workflows/ci-image.yaml [hunk 5]
```diff
@@ -0,0 +1,70 @@
+name: CI Image
```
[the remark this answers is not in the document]
iamrecursion: True, true. In that case, I think we should think a bit more about what the workflow looks like for someone actually doing that! 

Consider the scenario where I want to update the Rust version. For my PR's CI to run properly, it's going to need me to rebuild that container _before_ it can run. Some thoughts:

- Is there a way we can trigger the workflow based on a command from someone with write permissions or higher in the PR description? Or by pressing a button from that PR? If so, it should rebuild _before_ any of the other jobs execute, otherwise CI has the potential to fail spuriously.
- Can we have the workflows that run in the image run some check and fail out if they are expecting (e.g. via `rust-toolchain.toml` or `lean-toolchain`) an environment that differs from the installed one? This would allow us to have an informative error message if the updated image is _not_ available for the run.

Thoughts?

.github/workflows/ci-noir.yaml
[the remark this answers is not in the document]
iamrecursion: Unfortunately I don't think I have the permissions to do that, I don't think!

In terms of versioning, it's a little more complex. What we really care about for these containers is effectively the combination of the rust toolchain versions (stable, nightly), the Lean toolchain version, and the Noir version. Perhaps the version tag can be `rs<stable-version>-rn<nightly-version>-l<lean-version>-n<noir-version>` or similar? Currently this would resolve to something like (off the top of my head) `rs1.90.0-rn2025-07-27-lv4.21.0-...`.

iamrecursion:
When you want me to take a look at this again, please ping me and/or re-request review! Easiest way to get my attention!

I've also fixed the branch protection rules to look for the "Build CLI" job name instead of "Build Rust" so it shouldn't be blocked on the CI portion any more, just an approving review.

Eduardogbg:
a lot of comments I still need to address but at least now we have the image already pushed to `reilabs/lampe-ci:rust-1.90.0-lean-v4.21.0-noir-9a5b`

shame i didn't realize earlier i could have just built and pushed the image from this PR, I'll close #240 now, have already rebased this branch on top of those changes.

about your worries if this will still be faster after the image build step: yes, because we won't rebuild the image on every PR. and other changes about making the DAG better and hitting cache whenever possible go a long mile.

we have to decide on when we want to build the image but i think we can just run manually whenever we want to change the versions. requiring to merge something first to only then be able to update the image version could be annoying. but we can always check if the toolchain versions are all the same and if not, build and push a new image.

.github/workflows/ci-image.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build-image:
+    name: Build CI Image
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```
iamrecursion: I think we should be a bit more smart about this. Could we do something like the following?

1. Have this run every time.
2. Check if an image already exists for the right name (meaning the right tool versions).
3. If it exists, do nothing.
4. If it doesn't, build a new image?

That way we never have to deal with manually running the workflow if needed, and dependency dances for PRs (which make it very hard to rely on CI in such circumstances), and instead we just build the new image on-demand?

iamrecursion (commented):
This is really good stuff. A few more questions and comments about making this work smoothly!

.github/workflows/build-ci-image.yaml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/checkout@v6
+      - name: Read Toolchain Versions
+        id: versions
+        run: python3 scripts/ci/read_toolchain_versions.py
+      - name: Log In To GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
iamrecursion: I take it we don't need the `GHCR_TOKEN` now?

iamrecursion (approved):
This is fantastic! Thanks so much for all the hard work on this!

PR #240, opened 2026-01-07
Create CI job for building CI image

Eduardogbg:
We have to merge this before #239 so we can manually trigger the "Build CI Image" job, building the image to be used in the CI refactorings while relying solely on the `GITHUB_TOKEN` and not PATs


Eduardogbg:
managed to build `reilabs/lampe-ci:rust-1.90.0-lean-v4.21.0-noir-9a5b`. now it's up to us to decide when the image building happens. i think it can be triggered manually or on merge to main from now on?

PR #242, opened 2026-02-05
Fix caching across CI jobs

Eduardogbg:
done:
- remove examples, it's redundant with testing. maybe we can rename it to examples or something else

wip:
- make sure that everything that we build on one job is cached on the next, this includes the following lampe components:
  - rust cli
  - noir semantics in lean
  - noir stdlib formalization


iamrecursion:
The fact that there needs to be a CI cache bump manually like this indicates to me that we don't have a sufficiently detailed key for the cache?

Eduardogbg:
Yes correct, the cache key is not working correctly. I'm committing these to test. I will rebase the branch and remove a lot of the commits. Will ping you when it's ready for review again

PR #244, opened 2026-02-20
fix: extract comptime globals bound to simple identifiers

krushimir:
Comptime globals like `RATE: u32 = 3` were unconditionally skipped,
but the extracted code still referenced them, leaving dangling symbols.

Now we extract comptime globals that bind a simple identifier.

Found while proving `Poseidon2`: `RATE` was missing from extraction
but referenced in `new`, `perform_duplex`, and `absorb`.

### DIFF
--- a/src/lean/generator.rs
+++ b/src/lean/generator.rs
@@ -1380,12 +1380,15 @@ impl LeanGenerator<'_, '_, '_> {
         let global_data = self.context.def_interner.get_global(*id);
         let statement = self.context.def_interner.statement(&global_data.let_statement);
         let def_info = self.context.def_interner.definition(global_data.definition_id);
-        if def_info.comptime {
-            return None;
-        }
 
         match statement {
             HirStatement::Let(binding) => {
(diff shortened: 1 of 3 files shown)


src/lean/generator.rs
```diff
@@ -1380,12 +1380,15 @@ impl LeanGenerator<'_, '_, '_> {
...
-            return None;
-        }
 
         match statement {
             HirStatement::Let(binding) => {
+                // Skip comptime globals with non-identifier patterns (e.g. tuple
+                // destructuring) as they cannot be represented as named globals.
+                if def_info.comptime && !matches!(binding.pattern, HirPattern::Identifier(_)) {
+                    return None;
+                }
```
iamrecursion: Anything that actually has a constant value at extraction time _should_ be able to be extracted, so if we do see other valid globals I would file an issue to track extracting those.

iamrecursion (approved):
LGTM in general. Nice simple change.

PR #245, opened 2026-02-20
fix(elab): avoid ref reallocation for mutable locals

krushimir:
Found during Poseidon2 proving. When a mutable local (already a heap ref in Lampe)
is passed to `#_ref`, the Lean elaborator auto-dereferences it first, producing
`ref(readRef id)` - a fresh cell that loses reference identity and drops state updates.

A `suppressAutoDeref` flag in `DSLState` propagates through `makeExpr`'s recursive
tree walk, so `makeBareIdent` skips the `readRef` and the `#_ref` handler skips the
outer `ref(...)` - emitting `Expr.var id` directly. This handles the argument at any
nesting depth (bare ident, parens, blocks), not just top-level identifiers.

Regression tests prove a mutable-ref loop for all three cases.

### DIFF
--- a/Lampe/Lampe/Syntax/Builders.lean
+++ b/Lampe/Lampe/Syntax/Builders.lean
@@ -215,32 +215,39 @@ partial def makeExpr [MonadDSL m]
       (←``(Expr.callBuiltin _ $(←makeNoirType tp) $(←makeBuiltin name.getId.toString) $argVals))
       binder
       k
-  -- `makeExpr` auto-dereferences mutable locals, so `#_ref(expr)` would lower to `ref(readRef id)`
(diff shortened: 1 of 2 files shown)


iamrecursion (changes requested):
Generally a good start, but I have some changes I'd like to see inline.

I'm also concerned about how we would fail to handle chains of `ref(deref(ref(deref(...))))` properly when they are done explicitly. I don't know if noir can generate this code (even if a user writes `&(*(&(*(...))))` it could get collapsed) but not handling it feels like leaving a hole in our semantic model and hence making this fix incomplete.

iamrecursion (approved):
Looks good to me! Glad we're now handling the explicit chains of `&*&*&*...` for autoderef, but as we discussed on Slack please make sure it handles similar chain collapses where autoderef is not involved before merging.

PR #247, opened 2026-03-16
Update Noir Stdlib to v1.0.0-beta.14

Eduardogbg:
- requires updating the compiler to version v1.0.0-beta.19


Lampe/Lampe/Builtin/Slice.lean
```diff
@@ -118,4 +118,22 @@ def sliceRemove := newGenericPureBuiltin
...
+
+/-- Alias for `slicePopFront`. In newer Noir versions the builtin is called `vectorPopFront`. -/
+abbrev vectorPopFront := @slicePopFront
+
+/-- Alias for `sliceInsert`. In newer Noir versions the builtin is called `vectorInsert`. -/
+abbrev vectorInsert := @sliceInsert
+
+/-- Alias for `sliceRemove`. In newer Noir versions the builtin is called `vectorRemove`. -/
+abbrev vectorRemove := @sliceRemove
+
```
iamrecursion: It would make more sense to me to rename the actual definition to the new name, and then alias it under the _older_ name here. I would also add a comment noting the noir version where this change occurs.

iamrecursion (changes requested):
A few things, but mostly looks good. 

I've not checked in detail that the move of the namespace has left things unchanged except where relevant, so I'm definitely trusting that!

iamrecursion (changes requested):
Almost there! I think we want to handle the slice -> vector a little more gracefully, and can we also bump the Lampe library version as a result?

iamrecursion (approved):

PR #248, opened 2026-03-17
Test extraction and proving on real Noir libraries

krushimir:
Validates the end-to-end extraction and proving pipeline against real-world Noir libraries from the wild (https://github.com/noir-lang/awesome-noir).

### Integration tests added

- **[noir_sort](https://github.com/noir-lang/noir_sort)** - extraction + proofs for `sort_u32` and `sort_via_u32_sorted`, which uses `check_shuffle` as a nargo dependency.
- **[complexnr](https://github.com/manu-bhadoria/complexnr)** - extraction + proofs for complex number arithmetic.

### Extractor fixes (found during dogfooding)

Two bugs surfaced while extracting real-world libraries:
- **Nested kind annotation in `CheckedCast`** (`ast.rs`, `writer.rs`): Type-level arithmetic expressions containing a cast (e.g. `(N * 2 - 1) - 1`) generated a double `: u32` annotation, producing invalid Lean syntax. Fixed by propagating the cast's kind outward and emitting only the bare inner expression in `write_cast_type_expression`.
-  **Infix kind mismatch for generic operands** (`generator.rs`): Expressions like `N - 1` where one operand is a generic type variable (`Kind::Type`) and the other is a concrete numeric literal (`Kind::U(32)`) caused a panic on a kind equality assertion. Fixed by selecting the concrete kind when one side is `Kind::Type`.

Regression tests for both fixes are included in `src/lib.rs`.

### Simp lemmas

Added `@[simp]` lemmas for `Builtin.indexTpl` and `Builtin.replaceTuple` to
reduce boilerplate in struct field access proofs.


testing/Complexnr/lampe/complexnr-0.0.0.lean
```diff
@@ -0,0 +1,125 @@
...
+
+open Lampe
+
+set_option linter.unusedVariables false
+set_option autoImplicit false
+set_option maxRecDepth 4096
+
+namespace «complexnr-0.0.0»
+
+-- Lean-side type alias for Complex values: (real, imaginary, ())
```
iamrecursion:
```suggestion
```

PR #252, opened 2026-04-04
feat(stdlib): complete Map formalization surface -- core proved, all 18 method specs stated

MavenRain:


kustosz:
These are vacuous (`(fun _ => True)` postconditions) statements with no proofs

PR #253, opened 2026-04-04
feat(stdlib): crypto formalization -- Hash.Mod, Hash.Poseidon2, EmbeddedCurveOps specs and type projections

MavenRain:


kustosz:
Vacuous statements with no proofs

PR #256, opened 2026-04-16
fix: Use the crate of the struct in `generate_constructor`

### LINKED ISSUE #255: Code generated for structure literal uses wrong path for type
Take the following Noir source:
```noir
use std::embedded_curve_ops::EmbeddedCurveScalar;

pub fn main() {
    let _a = EmbeddedCurveScalar { hi: 1, lo: 2 };
    let _b = EmbeddedCurveScalar::new(3, 4);
}
```

When I generate Lean code from it using `lampe`, I get the following `Main.lean`:
```lean
-- Generated by lampe

import «fuzz_testing-0.0.0».Extracted.GeneratedTypes
import Lampe

open Lampe

noir_def «fuzz_testing-0.0.0»::main<>() -> Unit := {
  let _a = (#_makeData returning «fuzz_testing-0.0.0»::std::embedded_curve_ops::EmbeddedCurveScalar<>)((2: Field), (1: Field));
  let _b = («std-1.0.0-beta.14»::embedded_curve_ops::EmbeddedCurveScalar::new<> as λ(Field, Field) -> «std-1.0.0-beta.14»::embedded_curve_ops::EmbeddedCurveScalar<>)((3: Field), (4: Field));
  #_skip
}

def «fuzz_testing-0.0.0».Main.env : Env := Env.mk
  [«fuzz_testing-0.0.0::main»]
  []
```

Notice how:
* `_a` uses `«fuzz_testing-0.0.0»::std::embedded_curve_ops::EmbeddedCurveScalar` while 
* `_b` uses `«std-1.0.0-beta.14»::embedded_curve_ops::EmbeddedCurveScalar`. 

My `main.nr` happens to be in a package named `fuzz_testing`, and it looks like when we use the struct literal initialisation, it gets the name of the current project, whereas if we call a method on the type (such as `new`), the path is based on the top `use`. `std-1.0.0-beta.14` is the version of the Noir stdlib in Noir `1.0.0-beta.20`. 

When I try to build the Lean project with `lake build` I get the following error:
```
❯ lake build
...
Unpacked in 3080 ms
Completed successfully!
✖ [3062/3068] Building «fuzz_testing-0.0.0».Extracted.Main
trace: .> LEAN_PATH=...
error: fuzz_testing-0.0.0/Extracted/Main.lean:8:0: unknown identifier '«fuzz_testing-0.0.0::std::embedded_curve_ops::EmbeddedCurveScalar»'
error: Lean exited with code 1
Some required builds logged failures:
- «fuzz_testing-0.0.0».Extracted.Main
error: build failed
```

aakoshh:
Fixes https://github.com/reilabs/lampe/issues/255

Updated `TestDep` to have an example of this. When I first ran `./testing/test.xsh --test TestDep --update` to establish the baseline it was okay, but on subsequent commits failed with `git`, so I updated `Main.lean` manually 🤞

### DIFF
--- a/src/lean/generator.rs
+++ b/src/lean/generator.rs
@@ -24,138 +24,52 @@ use noirc_frontend::{
     },
     hir_def::{
         expr::{
-            HirArrayLiteral,
-            HirBlockExpression,
-            HirCallExpression,
-            HirCastExpression,
-            HirConstrainExpression,
-            HirConstructorExpression,
-            HirExpression,
-            HirIdent,
-            HirIfExpression,
-            HirIndexExpression,
-            HirInfixExpression,
-            HirLambda,
-            HirLiteral,
-            HirMemberAccess,
-            HirPrefixExpression,
+            HirArrayLiteral, HirBlockExpression, HirCallExpression, HirCastExpression,
+            HirConstrainExpression, HirConstructorExpression, HirExpression, HirIdent,
+            HirIfExpression, HirIndexExpression, HirInfixExpression, HirLambda, HirLiteral,
+            HirMemberAccess, HirPrefixExpression,
         },
         function::{FuncMeta, Param},
         stmt::{
-            HirAssignStatement,
-            HirForStatement,
-            HirLValue,
-            HirLetStatement,
-            HirPattern,
+            HirAssignStatement, HirForStatement, HirLValue, HirLetStatement, HirPattern,
             HirStatement,
         },
         traits::{NamedType, TraitConstraint, TraitImpl},
     },
     node_interner::{
-        DefinitionKind,
-        DependencyId,
-        ExprId,
-        FuncId,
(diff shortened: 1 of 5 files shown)


iamrecursion (approved):
Great catch! This was indeed a bug, and as you surmised in #255 it should be using the type's crate rather than the root crate!

aakoshh:
I think the E2E test is failing because the PR is from a fork:
```
error: /tmp/tmpllqs8z3n/lampe/../../../__w/_temp/lake-packages/GitDepWithLampe-2.0.0: revision not found 'af/fix-struct-path'
```

iamrecursion:
Alrighty. I'm going to force merge this, but if it goes red I'm going to back out the change on `main` and we'll need to investigate! Thanks for the contribution!

PR #280, opened 2026-04-29
Prove semantics for Poseidon2

Eduardogbg:
Defines semantics for Poseidon2 builtin, standard library functions, Hasher trait, and proves them.

Includes Noir tests for reference values of the hash, but since Noir's Poseidon2 instantiation is not used by anything else, the test is a bit circular, as I'm just hard-coding the observed output Noir's Poseidon2 has for a few reference strings. Serves more as a regression test. Ideally later we could cross-reference with an independent parametrizable implementation of Poseidon, instantiated to the same parameters (width, permute, field) as Noir.


Lampe/Lampe/Builtin/Stubs.lean
```diff
@@ -44,7 +45,15 @@ def fmtstrAsCtstring := stub
...
 def mkFormatString := stub
 def multiScalarMul := stub
-def poseidon2Permutation := stub
+/--
+Noir's `poseidon2_permutation` foreign builtin for the currently supported width-4 field array.
+
+The cryptographic permutation itself is modeled by `Crypto.Poseidon2.noirPermutation4`; the stdlib
+implements and proves the sponge layer around this builtin.
+-/
+def poseidon2Permutation := newTotalPureBuiltin
```
kustosz: move it out of the file called `Stubs` if it's not a stub

kustosz (changes requested):

kustosz (approved):

Issue #271, opened 2026-04-22
Associated constants: `Skipping global with invalid statement`


aakoshh:
The following Noir program can be successfully extracted by `lampe`, but it prints `Skipping global with invalid statement: Error` to the console:

```noir
pub trait Foo {
    let N: u32;
}

fn main() {}
```

This happens here: https://github.com/reilabs/lampe/blob/8d62d88e9d72a3dda69cb81f115c58989a9da25d/src/lean/generator.rs#L1384-L1390

As discussed in https://github.com/reilabs/lampe/issues/269#issuecomment-4282182769 associated constants are not supported by Lampe yet, but in this instance they are not rejected, because unlike in the examples in that comment, the constant here is not _used_. 

I investigated this in https://github.com/noir-lang/noir/issues/12392 and I think the fact that the `let_statement` is a `HirStatement::Error` is not invalid per-se; this global is for `let N: u32;` which doesn't have an actual value, and it's only global within the scope of the trait itself. The compiler creates a separate module for each trait in `ModCollector::collect_traits` and the `TraitItem::Constant` is defined as a global in that context. 

I couldn't see an easy way to tell that the `ModuleData` that Lampe is looking at in `generate_module_contents` is a trait (it doesn't have a flag like `is_type`), other than comparing its ID to all trait IDs. 

Just wanted to leave a note so that if in the future associated types are supported, you can choose to hide this message when the module is a trait, as this behaviour is expected at the moment (in Noir `1.0.0-beta.20`).

iamrecursion:
Great catch for this edge case where the existing code path is not exercised!

For the team member that works on this, see #272 for the general work item.

aakoshh:
https://github.com/noir-lang/noir/pull/12486 adds a new `HirStatement::TraitAssociatedConstant` for this, so you can skip the `eprintln!` in those cases.

PR #292, opened 2026-05-04
fix(generator): suppress invalid-statement warning for trait  associated constants (#271)

MavenRain:


kustosz:
Thnks, but we have a dedicated fix coming upstream, so this isn't needed: https://github.com/noir-lang/noir/pull/12486

PR #297, opened 2026-05-21
Bump Lean and Mathlib to v4.29.1

Eduardogbg:
- Bumps the Lean toolchain and Mathlib to v4.22.0.
- v4.22 makes Mathlib's elliptic-curve `+` computable (upstream [PR #27299](https://github.com/leanprover-community/mathlib4/pull/27299)), which the follow-up embedded-curve and ECDSA work relies on.
- v4.22 adds an `unusedSimpArgs` linter; the follow-up commit drops the stale simp args it flagged in the files modified on this branch.


iamrecursion:
Let me know when this is ready for review.

Eduardogbg:
need to fix CI then I'll investigate whether we can bump further. honestly think this will be done by tomorrow most likely, having to wait for CI etc takes some time

PR #306, opened 2026-06-05
Formalize Pedersen hash + commitment using BLAKE3 generators

Eduardogbg:
> ⚠️ Stack order: `main` ← `edu/blake3` ← `edu/bn254-refactor` ← `edu/pedersen`.
> Merge upwards in that order; this PR retargets to `main` once the intermediates land.

Adds the Lampe semantics for Noir's `pedersen_hash` and
`pedersen_commitment` (with and without separator).

Honestly this was the trickiest builtin to model, still not completely sure I like how the specs turned out. I tried to ground everything on what the Barretenberg gadget's actually constrain. Without some canonicality guarantees a lot of the specs required some existential statements for the limbs decomposition around MSM. Had to change MSM specs because of this, but I think our model is more precise now, it does appear it guarantees canonical decomposition (source: `cycle_group::batch_mul` → `straus_scalar_slice.cpp:59`, `create_limbed_range_constraint`).

Had to implement a fast square-root algorithm for prime fields, landed on Tonelli-Shanks since it we can fix a quadratic non-residue independently of input (not the case for Cipolla's).

### DIFF
--- a/Lampe/Lampe/Builtin/Crypto/EmbeddedCurve.lean
+++ b/Lampe/Lampe/Builtin/Crypto/EmbeddedCurve.lean
@@ -20,18 +20,32 @@ def embeddedCurveAdd := newGenericPureBuiltin
         ⟨[encodeCurvePoint ((curvePoint? p1).get h1 + (curvePoint? p2).get h2)], by simp⟩⟩)
 
 /--
-Noir's `multi_scalar_mul_array_return` foreign builtin. The result is
-`∑ᵢ (Scalar.valueNat sᵢ) • Pᵢ` computed via Mathlib's `+` and `nsmul`.
-On-curve obligation is a precondition: every input point must lift
-through `curvePoint?`.
+Noir's `multi_scalar_mul_array_return` foreign builtin.
+
+Two preconditions, both modelling actual circuit constraints emitted by
+the Barretenberg MSM gadget:
+
+- **On-curve**: every input point lifts through `curvePoint?`. The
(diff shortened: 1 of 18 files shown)


Lampe/Lampe/Builtin/Crypto/EmbeddedCurve.lean
```diff
@@ -20,18 +20,45 @@ def embeddedCurveAdd := newGenericPureBuiltin
...
+
+- **On-curve**: every input point lifts through `curvePoint?`. The
+  gadget's curve-relation constraint inside `cycle_group::batch_mul`
+  enforces this on each input.
+- **Canonical scalars**: every input scalar satisfies `scalarCanonical`
+  (`lo.val < 2^128 ∧ hi.val < 2^126`). The gadget's
+  `create_limbed_range_constraint` emits this on every input scalar via
+  `straus_scalar_slice.cpp:59`, with `LO_BITS`/`HI_BITS` pinned in
+  `cycle_scalar.hpp:38-44`. Receipts in `docs/msm-gadget-receipts.md`
+  (local note, not upstreamed).
```
kustosz: That local note is a  rather unhelpful reference.

kustosz (approved):
Approving, but please fix the wrong doc

Lampe/Lampe/Builtin/Crypto/EmbeddedCurve.lean
Eduardogbg: oops sorry, slipped past me

PR #307, opened 2026-06-11
Bake Mathlib build cache into CI image

Eduardogbg:
Mathlib cache migrated to Azure and they stopped serving our version's build cache. That and a couple of other reasons (e.g. cache get having a few types of transient errors, github itself being an unreliable mess) makes our CI pipelines very finicky, especially when you consider how long a cacheless Mathlib build takes.

This PR bakes the Mathlib build cache on our CI image. Since we rebuild the image whenever Mathlib (among other core dependencies) changes version, this has basically no downside.

A couple of other changes here like:
- having another free space step, since regardless of our image, non-self-hosted Github Actions runners include a bunch of unwanted software like Android SDK which occupies a lot of disk space
- provides lake manifests for the test packages


.github/workflows/build-ci-image.yaml
```diff
@@ -35,6 +35,12 @@ jobs:
         id: check
         continue-on-error: true
         run: python3 scripts/ci/verify_ci_image.py ghcr.io/reilabs/lampe-ci:${{ steps.versions.outputs.image_tag }}
```
piohei: What is the reason to have it as a script instead of just using docker command to check?

.github/workflows/build-ci-image.yaml
```diff
@@ -35,6 +35,12 @@ jobs:
...
+      - name: Free Runner Disk Space
+        if: steps.check.outcome == 'failure' && inputs.allow_push
```
piohei: What about always removing it if don't need it at all? is the step very slow and it's better to do it only when in need to rebuild the image?

piohei: Also all other steps share same condition that could be a condition for another job in the pipeline (like run build image job only if there is no image already build).
https://github.com/orgs/community/discussions/25941

.github/workflows/ci.yaml
```diff
@@ -174,28 +173,23 @@ jobs:
...
           path: Lampe/.lake/build
-          key: lampe-build-${{ runner.os }}-${{ hashFiles('Lampe/**/*.lean', 'Lampe/**/*.toml', 'Lampe/**/*.json', 'Lampe/lean-toolchain', 'Lampe/lake-manifest.json', 'Lampe/lakefile.toml') }}
+          key: lampe-build-${{ runner.os }}-${{ hashFiles('Lampe/**/*.lean', 'Lampe/**/*.toml', 'Lampe/**/*.json', 'Lampe/lean-toolchain', 'Lampe/lake-manifest.json') }}
           restore-keys: |
             lampe-build-${{ runner.os }}-
       - name: Get Mathlib Cache
         if: ${{ steps.lampe-build-cache.outputs.cache-hit != 'true' }}
         working-directory: Lampe
-        run: lake exe cache get
+        run: lake exe cache get || echo "cache get failed; falling back to artifacts baked into the CI image"
```
piohei: Why not always use baked in docker image files?

.github/workflows/build-ci-image.yaml
Eduardogbg: good catch. changed it to just a docker command on ci

.github/workflows/ci.yaml
Eduardogbg: this command reads the local archive (via an env var pointing to it) and unpackages. it shouldn't hit the network

.github/workflows/build-ci-image.yaml
Eduardogbg: ok good suggestion: build-ci-image is now check/build/guard jobs with job-level conditions, and the disk-space step only exists inside the build job, so it only runs when an image actually gets built. did the same with ci.yaml, the per-step cache-hit guards are gone

piohei (approved):
good work!

PR #315, opened 2026-09-17
fix: require the postcondition at the final state in `Omni.loopDone`

### LINKED ISSUE #314: `Omni.loopDone` proves any postcondition for zero-iteration loops, including `False`
*(This issue was written by AI tooling)*

`Lampe/Lampe/Semantics.lean:84-86` declares the `loopDone` rule of the omni-semantics with no premise about the postcondition `Q`, so any loop with `lo ≥ hi` satisfies every postcondition, including one that rejects every outcome. Through the `letIn` rule, an empty loop in sequence position also discharges the proof obligation for any continuation without checking that continuation.

The rule reads:

```lean
| loopDone :
  lo ≥ hi →
  Omni Γ st (.loop lo hi body) Q
```

The value-producing terminal rules (literals, constants, `var`, `skip`) all require `Q (some (st, v))` for their final state, and an empty loop should behave like `skip`. `loopDone` alone derives `Omni Γ st e Q` without requiring anything of `Q`. Here is the shape with no imports (I checked it on Lean v4.33.1):

```lean
inductive Judgment : (lo hi : Nat) → (post : Unit → Prop) → Prop
  | loopDone : lo ≥ hi → Judgment lo hi post

example (post : Unit → Prop) : Judgment 0 0 post := .loopDone (Nat.le_refl 0)
example : Judgment 0 0 (fun _ => False) := .loopDone (Nat.le_refl 0)
```

The tested revision is `919a38712d22d05fc83e59af17ef57077abda1e0`, the current `main`, with the pinned `leanprover/lean4:v4.29.1` toolchain and the checked-in manifest, on Apple Silicon.

## Reproduction

```sh
git clone https://github.com/reilabs/lampe.git
cd lampe
git checkout 919a38712d22d05fc83e59af17ef57077abda1e0
cd Lampe
lake exe cache get
lake build Lampe
```

Save this as `LoopAudit.lean` in the repository root:

```lean
import Lampe.Semantics
import Lampe.Hoare.SepTotal

open Lampe

namespace LoopAudit

theorem empty_loop_arbitrary_post (p : Prime) (Γ : Env) (st : State p)
    (Q : Option (State p × Unit) → Prop) :
    Omni p Γ st (.loop (0 : U 32) 0 (fun _ => .skip)) Q := by
  apply Omni.loopDone
  simp

theorem empty_loop_false_post (p : Prime) (Γ : Env) :
    STHoare p Γ ⟦⟧ (.loop (0 : U 32) 0 (fun _ => .skip)) (fun _ => ⟦False⟧) := by
  intro H st _
  apply Omni.loopDone
  simp

theorem program_after_empty_loop_arbitrary_post (p : Prime) (Γ : Env) (st : State p)
    {tp : Tp} (k : Tp.denote p .unit → Expr (Tp.denote p) tp)
    (Q : Option (State p × Tp.denote p tp) → Prop) :
    Omni p Γ st (.letIn (.loop (0 : U 32) 0 (fun _ => .skip)) k) Q := by
  apply Omni.letIn (Q₁ := fun _ => False)
  · apply Omni.loopDone
    simp
  · intro _ _ h
    exact h.elim
  · intro h
    exact h.elim

#print axioms empty_loop_arbitrary_post
#print axioms empty_loop_false_post
#print axioms program_after_empty_loop_arbitrary_post

end LoopAudit
```

Run `lake env lean ../LoopAudit.lean`. It exits 0, and each theorem reports only the standard axioms:

```text
'LoopAudit.empty_loop_arbitrary_post' depends on axioms: [propext, Classical.choice, Quot.sound]
'LoopAudit.empty_loop_false_post' depends on axioms: [propext, Classical.choice, Quot.sound]
'LoopAudit.program_after_empty_loop_arbitrary_post' depends on axioms: [propext, Classical.choice, Quot.sound]
```

Machine-checked versus read: the shape snippet above and the three theorems are machine-checked (the snippet on bare Lean v4.33.1, the theorems at the pinned toolchain). The claims about extracted `for` loops and the standard-library proof layer come from reading those sources; I did not rebuild the `stdlib/` or `testing/` projects.

## Why it matters

In `Omni`, the postcondition is a predicate over the outcome: `some` for successful evaluation, `none` for failure. `empty_loop_arbitrary_post` instantiated at `Q := fun _ => False` gives a judgment that rejects both success and failure, which no expression should satisfy. At the Hoare level, `STHoare` allows failure and requires its postcondition only after successful evaluation, so a `False` postcondition alone does not imply inconsistency; here, however, the expression is an empty loop, which should return unit without changing the state, and the `Omni`-level instantiation above rejects the failure reading as well.

The third theorem shows how it spreads: an empty loop in sequence position discharges the continuation's proof obligation without checking the continuation, because the intermediate postcondition of `letIn` can be `fun _ => False`. Any Noir `for lo..hi` loop with `lo ≥ hi` extracts to such a loop. The standard-library proof layer contains loops with generic bounds, for example `Expr.loop (↑0) MaxLen` in `stdlib/lampe/Stdlib/Collections/BoundedVec/HigherOrderMap.lean:142`, where `MaxLen : U 32` is unrestricted, so `MaxLen = 0` is a legal instantiation.

## Fix

One added premise, mirroring `skip`:

```diff
 | loopDone :
   lo ≥ hi →
+  Q (some (st, ())) →
   Omni Γ st (.loop lo hi body) Q
```

The patch changes two existing proof bodies to discharge the new premise: the `loopDone` case of `Omni.frame` in `Lampe/Lampe/Semantics.lean` and `STHoare.loopDone_intro` in `Lampe/Lampe/Hoare/SepTotal.lean` (via `SLP.ent_star_top`). No handwritten theorem statement changes; the constructor gains a premise, so its generated recursor changes with it. The statements of `STHoare.loopDone_intro`, `loopNext_intro`, and both loop invariant rules are untouched, so proofs built on the `STHoare` interface, like the Base64 proofs under `testing/`, keep elaborating against the same statements.

I verified the patch at the same revision and toolchain. `lake build Lampe` and `lake build Tests` succeed with zero diagnostics both before and after it, and all 21 test modules build. With the patch applied, the three theorems above all fail to elaborate: the first two are stuck on the new `Q (some (st, ()))` obligation, the third on its intermediate `False`. The empty loop still proves its intended specification, `Omni p Γ st (.loop (0 : U 32) 0 (fun _ => .skip)) (fun v => v = some (st, ()))`, and `#print axioms` on that proof, on `Omni.frame`, `Omni.consequence`, `STHoare.loopDone_intro`, `loopNext_intro`, and both loop invariant rules reports only `[propext, Classical.choice, Quot.sound]`.

I am happy to open a PR with the patch and a regression test if you prefer that route. Lampe is a great tool and we would like to build on it, which is how we ran into this.

oskarth:
Fixes #314.

The `loopDone` rule concluded the `Omni` judgment for an arbitrary postcondition from the bounds check alone, so a zero-iteration loop satisfied every postcondition, and through `letIn` it discharged any continuation's obligation unchecked. Details and reproduction are in #314.

This adds the premise `Q (some (st, ()))`, mirroring `skip`, and repairs the two proofs that build the rule directly: the `loopDone` case of `Omni.frame` and `STHoare.loopDone_intro`. No handwritten theorem statement changes. `Tests/LoopDone.lean` adds five regression theorems; the existing `Tests` glob discovers the module.

Verified at the pinned `leanprover/lean4:v4.29.1` toolchain on top of `919a387`: `lake build Lampe` and `lake build Tests` are clean before and after, and the exploit proofs from #314 no longer elaborate.


Lampe/Tests/LoopDone.lean
```diff
@@ -0,0 +1,49 @@
...
+  intro h
+  have hrun := h ⟦⟧ ∅ (by simp [SLP.lift])
+  have hpost := (empty_loop_iff p Γ ∅ _ (by simp)).mp hrun
+  simp [SLP.star, SLP.lift] at hpost
+
+#print axioms empty_loop_iff
+#print axioms empty_loop_preserves_state
+#print axioms empty_loop_rejects_false
+#print axioms empty_loop_triple
+#print axioms empty_loop_rejects_false_triple
```
iamrecursion: These don't check anything, so I think they can be removed.

Lampe/Tests/LoopDone.lean
```diff
@@ -0,0 +1,49 @@
...
+    ¬ Omni p Γ st (.loop lo hi body) (fun _ => False) := by
+  intro h
+  exact (empty_loop_iff p Γ st _ hle).mp h
+
+theorem empty_loop_triple (p : Prime) (Γ : Env) (P : SLP (State p))
+    {lo hi : U s} {body : U s → Expr (Tp.denote p) tp} (hle : lo ≥ hi) :
+    STHoare p Γ P (.loop lo hi body) (fun _ => P) := by
+  intro H st hp
+  apply Omni.loopDone hle
+  exact SLP.ent_star_top st hp
```
iamrecursion: These all fall out directly, so it is probably worth a comment that `empty_loop_iff` and `empty_loop_rejects_false_triple` are sufficient to pin the fixed behaviour. I would keep these three for documentation clarity, however.

Lampe/Tests/LoopDone.lean
```diff
@@ -0,0 +1,49 @@
...
+
+theorem empty_loop_rejects_false_triple (p : Prime) (Γ : Env) :
+    ¬ STHoare p Γ ⟦⟧ (.loop (0 : U 32) 0 (fun _ => .skip)) (fun _ => ⟦False⟧) := by
```
iamrecursion: This should take `hle` the same way as its siblings and be general over any `lo ≥ hi`.

Lampe/Tests/LoopDone.lean
```diff
@@ -0,0 +1,49 @@
...
+    Omni p Γ st (.loop lo hi body) (fun result => result = some (st, ())) := by
+  exact Omni.loopDone hle rfl
+
+theorem empty_loop_rejects_false (p : Prime) (Γ : Env) (st : State p)
+    {lo hi : U s} {body : U s → Expr (Tp.denote p) tp} (hle : lo ≥ hi) :
+    ¬ Omni p Γ st (.loop lo hi body) (fun _ => False) := by
+  intro h
+  exact (empty_loop_iff p Γ st _ hle).mp h
+
+theorem empty_loop_triple (p : Prime) (Γ : Env) (P : SLP (State p))
```
iamrecursion: This honestly probably makes more sense as a library lemma. It is strictly stronger than `loopDone_intro` with (`⟦⟧`, `lo = lo`) so I suggest moving it next to that so others can benefit.

iamrecursion (commented):
Thanks so much for your contribution!

Fundamentally the change is sound. What I've suggested here is some minor cleanups more than anything!

